PR #2, opened 2017-05-19
Implement importer using ChangeSet.

jortel:
Pilot Implementation of the importer using ChangeSet

Requires: https://github.com/pulp/pulp/pull/3006

### DIFF
--- a/pulp_file/app/models.py
+++ b/pulp_file/app/models.py
@@ -1,26 +1,65 @@
+import os
+
+from collections import namedtuple
+from gettext import gettext as _
+from logging import getLogger
+from urllib.parse import urlparse, urlunparse
+
 from django.db import models
 
-from pulpcore.plugin import models as platform
+from pulpcore.plugin.models import Artifact, Content, Importer, Publisher
+from pulpcore.plugin.changeset import (
+    BatchIterator, ChangeSet, SizedIterable, RemoteContent, RemoteArtifact,
+    ChangeReport, ChangeFailed)
+
+from pulp_file.manifest import Manifest
 
 
-class FileContent(platform.Content):
+log = getLogger(__name__)
+
+
+# Changes needed.
+Delta = namedtuple('Delta', ('additions', 'removals'))
+# Natural key.
+Key = namedtuple('Key', ('path', 'digest'))
+
+
+class FileContent(Content):
     """
     The "file" content type.
 
-    Content of this type represents a collection of 0 or more files identified by a unique name.
+    Content of this type represents a collection of 0 or more files uniquely
+    identified by path and SHA256 digest.
 
     Fields:
-        name (str): A unique name for this content.
+        path (str): The file relative path.
+        digest (str): The SHA256 HEX digest.
 
     """
     TYPE = 'file'
 
-    name = models.TextField(unique=True, null=False)
+    path = models.TextField(blank=False, null=False)
+    digest = models.TextField(blank=False, null=False)
 
-    natural_key_fields = (name,)
+    natural_key_fields = (path, digest)
+
+    class Meta:
+        unique_together = (
+            'path',
+            'digest'
+        )
+
+    def natural_key(self):
+        """
+        Get the model's natural key.
+
+        Returns:
+            Key: The natural key.
+        """
+        return Key(path=self.path, digest=self.digest)
 
 
-class FileImporter(platform.Importer):
+class FileImporter(Importer):
     """
     Importer for "file" content.
     """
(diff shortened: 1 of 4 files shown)


pulp_file/app/models.py
```diff
@@ -20,22 +37,124 @@ class FileContent(platform.Content):
...
+            manifest (Manifest): The fetched manifest.
+            inventory (set): Set of existing content (natural) keys.
+            mirror (bool): Faked mirror option.
+                TODO: should be replaced with something standard.
+
+        Returns:
+            Delta: The needed changes.
+        """
+        delta = Delta(
+            additions={e.path for e in manifest.read()},
```
mhrivnak: As long as you're using sets, how about some set math!

```
remote_content = {e.path for e in manifest.read()}
delta.additions = remote_content - inventory
if mirror:
  delta.removals = inventory - remote_content
return delta
```

mhrivnak:
FWIW, I think the uniqueness constraint (unit key) needs to include more than just the name, to accommodate two files with the same name in different repos. In Pulp 2, the unit key was (name, checksum, size). In theory, name and checksum should be sufficient.

mhrivnak:
oops, I didn't mean to close it. I just hit the wrong button. :)

bmbouter:
@jortel I hope this is also merged along with the changeset PR. I haven't reviewed this closely, so I won't leave a formal lgtm, but I do hope it gets merged soon.

pulp_file/app/models.py
```diff
@@ -1,41 +1,230 @@
...
 
-        Sync behavior for the file plugin has not yet been implemented.
+        Returns:
+            set: of Key.
         """
-        raise NotImplementedError
+        inventory = set()
+        q_set = FileContent.objects.filter(repositories=self.repository)
+        q_set = q_set.only(*[f.name for f in FileContent.natural_key_fields])
+        for content in (c.cast() for c in q_set):
```
asmacdo: Will this try to keep all the units in memory or is there some fancy generator magic under the hood?

pulp_file/app/models.py
```diff
@@ -1,41 +1,230 @@
...
+        Generate the content to be added.
+
+        Args:
+            manifest (Manifest): The fetched manifest.
+            delta (Delta): The needed changes.
+
+        Returns:
+            generator: A generator of content to be added.
+        """
+        parsed_url = list(urlparse(self.feed_url))
```
asmacdo: I think you did it this way because urlparse().path is immutable. It would be more readable to use the `_replace` method of the namedtuple. In this case, the underscore is to prevent collisions, not to indicate a private method.

https://docs.python.org/3/library/collections.html#collections.namedtuple

    parsed_url = urlparse(self.feed_url)
    parsed_url._replace(path=new_path)

pulp_file/app/models.py
```diff
@@ -1,41 +1,230 @@
...
+        Returns:
+            generator: A generator of content to be removed.
+
+        """
+        for natural_keys in BatchIterator(delta.removals):
+            q = models.Q()
+            for key in natural_keys:
+                q |= models.Q(filecontent__path=key.path, filecontent__digest=key.digest)
+            q_set = self.repository.content.filter(q)
+            q_set.only('artifacts')
```
asmacdo: q_set = q_set.only('artifacts')

pulp_file/app/models.py
jortel: generator magic.

pulp_file/app/models.py
jortel: Oh, interesting (and weird they used _ prefix instead of _ suffix to avoid collisions).  But good to know.  Much simpler and will change it.

pulp_file/app/models.py
jortel: good catch!

pulp_file/manifest.py
```diff
@@ -0,0 +1,81 @@
...
+
+        Returns:
+            Entry: An entry.
+
+        Raises:
+            ValueError: on parsing error.
+        """
+        part = [s.strip() for s in line.split(',')]
+        if len(part) != 3:
+            raise ValueError(_('Must be: <path>, <size>, <digest>'))
```
jortel: How about I include the offending line#?

pulp_file/app/models.py
```diff
@@ -1,41 +1,239 @@
...
+                    removed += 1
+
+        log.warn(
+            _('FileImporter: total: added:%(a)d, removed:%(r)d, failed %(f)d'),
+            {
+                'a': added,
+                'r': removed,
+                'f': failed
+            })
+        # On failed > 0, raise a PulpCodedException?
```
asmacdo: I don't we should raise but lets remove the comment for now.

jortel: This PR is intended to be a starting point for the plugin and I expect it to get revised as we evolve the plugin API etc.  I would like to leave the comment in for now.

pulp_file/app/models.py
```diff
@@ -1,41 +1,239 @@
...
+                report.result()
+            except ChangeFailed:
+                failed += 1
+            else:
+                if report.action == ChangeReport.ADDED:
+                    added += 1
+                else:
+                    removed += 1
+
+        log.warn(
```
asmacdo: I think info is the right log level for this.

pulp_file/app/models.py
```diff
@@ -1,41 +1,239 @@
...
+    def sync(self):
+        """
+        TODO: rename to synchronize()
+        Synchronize the repository with the remote.
+        """
+        failed = 0
+        added = 0
+        removed = 0
+
+        log.warn(
```
asmacdo: How about debug level?

pulp_file/app/models.py
jortel: Agreed.  logging at WARN for now because something re: logging is not yet configured properly for workers in pulp3.  I added a TODO: to change this later.

PR #5, opened 2017-07-13
Improve PyPI Packaging

werwty:
re #2842
https://pulp.plan.io/issues/2842

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -8,12 +8,13 @@ requirements = [
 
 setup(
     name='pulp-file',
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -8,12 +8,13 @@
 
 setup(
     name='pulp-file',
-    version='0.0.1a1.dev0',
+    version='0.0.1a1.dev1',
```
bmbouter: I didn't expect this to bump but it's all pre-release stuff so that is fine.

PR #7, opened 2017-07-21
Add publisher.

jortel:
https://pulp.plan.io/issues/2925

Initial add of publisher.

Requires: https://github.com/pulp/pulp/pull/3100


pulp_file/app/models.py
```diff
@@ -254,4 +263,49 @@ def publish(self):
 
         Publish behavior for the file plugin has not yet been implemented.
```
dkliban: Updated docstring needed

jortel:
@dkliban, tested against rebased core.

PR #8, opened 2017-07-29
Problem: model changes in pulpcore broke FileImporter

dkliban:
Solution: add BasicFileImporter that uses new models from pulpcore

Also introduces a PulpManifestMixin that provides functions for downloading a
PULP_MANIFEST and calculating a delta for repository content. This mixin is
used by both the FileImporter and BasicFileImporter.

re #2872
https://pulp.plan.io/issues/2872

### DIFF
--- a/pulp_file/app/models.py
+++ b/pulp_file/app/models.py
@@ -1,3 +1,4 @@
+import hashlib
 import os
 
 from collections import namedtuple
@@ -5,9 +6,14 @@ from gettext import gettext as _
 from logging import getLogger
 from urllib.parse import urlparse, urlunparse
 
+from django.core.files import File
 from django.db import models
+from django.db.utils import IntegrityError
 
-from pulpcore.plugin.models import Artifact, Content, Importer, Publisher
+from pulpcore.download import DigestValidation, DownloadError, SizeValidation
+from pulpcore.plugin.tasking import Task as current_task
+from pulpcore.plugin.models import (Artifact, Content, ContentArtifact, DeferredArtifact, Importer,
+    ProgressBar, Publisher, RepositoryContent)
 from pulpcore.plugin.changeset import (
     BatchIterator, ChangeSet, SizedIterable, RemoteContent, RemoteArtifact,
     ChangeReport, ChangeFailed)
@@ -17,11 +23,12 @@ from pulp_file.manifest import Manifest
 
 log = getLogger(__name__)
 
+BUFFER_SIZE = 65536
 
 # Changes needed.
 Delta = namedtuple('Delta', ('additions', 'removals'))
 # Natural key.
-Key = namedtuple('Key', ('path', 'digest'))
+FileTuple = namedtuple('Key', ('path', 'digest', 'size'))
 
 
 class FileContent(Content):
@@ -59,9 +66,72 @@ class FileContent(Content):
         return Key(path=self.path, digest=self.digest)
 
 
-class FileImporter(Importer):
+class PulpManifestMixin(object):
     """
-    Importer for "file" content.
+    Provides functions for working with an importer's PULP_MANIFEST.
+    """
+    def _fetch_inventory(self):
+        """
+        Fetch existing content in the repository.
+
+        Returns:
+            set: of Key.
+        """
+        inventory = set()
+        q_set = FileContent.objects.filter(repositories=self.repository)
+        q_set = q_set.only(*[f.name for f in FileContent.natural_key_fields])
+        for content in (c.cast() for c in q_set):
+            key = FileTuple(path=content.path, digest=content.digest,size=content.size)
+            inventory.add(key)
+        return inventory
+
+    def _fetch_manifest(self):
+        """
+        Fetch (download) the manifest.
+
+        Returns:
+            Manifest: The manifest.
+        """
+        parsed_url = urlparse(self.feed_url)
+        download = self.get_download(self.feed_url, os.path.basename(parsed_url.path))
+        download()
+        return Manifest(download.writer.path)
+
+    @staticmethod
+    def _find_delta(manifest, inventory, mirror=True):
+        """
+        Using the manifest and set of existing (natural) keys,
+        determine the set of content to be added and deleted from the
+        repository.  Expressed in natural key.
+        Args:
+            manifest (Manifest): The fetched manifest.
+            inventory (set): Set of existing content (natural) keys.
+            mirror (bool): Faked mirror option.
+                TODO: should be replaced with something standard.
+
+        Returns:
+            Delta: The needed changes.
+        """
+        remote = set()
+        for entry in manifest.read():
+            key = FileTuple(path=entry.path, digest=entry.digest, size=entry.size)
+            remote.add(key)
+        additions = remote - inventory
+        if mirror:
+            removals = inventory - remote
+        else:
+            removals = set()
+        return Delta(additions=additions, removals=removals)
+
+
+
+class FileImporter(Importer, PulpManifestMixin):
+    """
+    Importer for "file" content that uses ChangeSets to perform the sync.
+
+    This importer uses the ChangeSet API to download content in parallel and import into the
+    repository. Any content removed from the remote repository since the previous sync is also
+    removed from the Pulp repository.
     """
     TYPE = 'file'
 
@@ -104,59 +174,6 @@ class FileImporter(Importer):
         # On failed > 0, raise a PulpCodedException?
         # Done
 
-    def _fetch_inventory(self):
-        """
-        Fetch existing content in the repository.
-
-        Returns:
-            set: of Key.
-        """
-        inventory = set()
-        q_set = FileContent.objects.filter(repositories=self.repository)
-        q_set = q_set.only(*[f.name for f in FileContent.natural_key_fields])
-        for content in (c.cast() for c in q_set):
-            key = Key(path=content.path, digest=content.digest)
-            inventory.add(key)
-        return inventory
-
-    def _fetch_manifest(self):
-        """
-        Fetch (download) the manifest.
-
-        Returns:
-            Manifest: The manifest.
-        """
-        parsed_url = urlparse(self.feed_url)
-        download = self.get_download(self.feed_url, os.path.basename(parsed_url.path))
-        download()
-        return Manifest(download.writer.path)
-
-    @staticmethod
-    def _find_delta(manifest, inventory, mirror=True):
-        """
-        Using the manifest and set of existing (natural) keys,
-        determine the set of content to be added and deleted from the
-        repository.  Expressed in natural key.
-        Args:
-            manifest (Manifest): The fetched manifest.
-            inventory (set): Set of existing content (natural) keys.
-            mirror (bool): Faked mirror option.
-                TODO: should be replaced with something standard.
-
-        Returns:
-            Delta: The needed changes.
-        """
-        remote = set()
-        for entry in manifest.read():
-            key = Key(path=entry.path, digest=entry.digest)
-            remote.add(key)
-        additions = remote - inventory
-        if mirror:
-            removals = inventory - remote
-        else:
-            removals = set()
-        return Delta(additions=additions, removals=removals)
-
     def _build_additions(self, manifest, delta):
         """
         Generate the content to be added.
@@ -171,7 +188,7 @@ class FileImporter(Importer):
         parsed_url = urlparse(self.feed_url)
         root_dir = os.path.dirname(parsed_url.path)
         for entry in manifest.read():
-            key = Key(path=entry.path, digest=entry.digest)
+            key = FileTuple(path=entry.path, digest=entry.digest, size=entry.size)
             if key not in delta.additions:
                 continue
             path = os.path.join(root_dir, entry.path)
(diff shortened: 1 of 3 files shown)


pulp_file/app/models.py
```diff
@@ -248,3 +260,136 @@ def publish(self):
...
+                    continue
+                path = os.path.join(root_dir, entry.path)
+                url = urlunparse(parsed_url._replace(path=path))
+                content, created = FileContent.objects.get_or_create(path=entry.path,
+                                                                     digest=entry.digest)
+                # Add content to the repository
+                association = RepositoryContent(
+                    repository=self.repository,
+                    content=content)
+                association.save()
```
goosemania: Is there any reason to create an `association` here, in the beginning, and not after everything is resolved with artifacts?
If there is some exception with artifacts creation, `association` needs to be cleaned up. If we  add content to a repo later, there will be no need for that clean up.

pulp_file/app/models.py
```diff
@@ -248,3 +260,136 @@ def publish(self):
...
+                # ContentArtifact is required for all download policies. It is used for publishing.
+                content_artifact = ContentArtifact(content=content,
+                                                   relative_path=entry.path)
+                # DeferredArtifact is required for deferred download policies.
+                deferred_artifact = DeferredArtifact(url=url, importer=self,
+                                                     content_artifact=content_artifact,
+                                                     sha256=entry.digest)
+
+                if self.download_policy != self.IMMEDIATE:
```
goosemania: Maybe it will be more readable to put handling of immediate case and lazy one into separate private methods?

pulp_file/app/models.py
```diff
@@ -248,3 +260,136 @@ def publish(self):
...
+                    # Set Artifact ID if the Artifact already exists and then continue
+                    try:
+                        artifact = Artifact.objects.get(sha256=entry.digest)
+                        content_artifact.artifact = artifact
+                    except Artifact.DoesNotExist:
+                        pass
+                    # Try to save ContentArtifact which may already exist
+                    try:
+                        content_artifact.save()
+                    except IntegrityError:
```
goosemania: I suggest to use [get_or_create](https://docs.djangoproject.com/en/1.11/ref/models/querysets/#get-or-create) here.
And set `deferred_artifact.content_artifact` here, becasue right now `deferred_artifact` is not updated if `content_artifact` already exists.

How about creating `ContentArtifact` and `DeferredArtifact` right before they are used in code and not in the beginnign of this `sync` method? I think that will decrease chances to make a bug if there is a possibility that object will be changed, like here.

pulp_file/app/models.py
```diff
@@ -248,3 +260,136 @@ def publish(self):
...
+                        content.delete()
+                        # TODO: record non-fatal exception in the Task
+                        continue
+                    else:
+                        with File(open(download.writer.path, mode='rb')) as file:
+                            checksums = self.get_checksums(file)
+                            try:
+                                artifact = Artifact(file=file, size=file.size, **checksums)
+                                artifact.save()
+                            except IntegrityError:
```
goosemania: In case of any other issue with saving artifact, `content` and `association` are not cleaned up.

Artifact save is a special case, because it's not only a database write, it's also works with a file storage. So any issues with it, can raise an exception here. E.g. I had wrong permissions for `/var/lib/pulp/`.

pulp_file/app/models.py
```diff
@@ -248,3 +260,136 @@ def publish(self):
...
+                                artifact = Artifact.objects.get(sha256=checksums['sha256'])
+                content_artifact.artifact = artifact
+                # Try to save ContentArtifact which may already exist
+                try:
+                    content_artifact.save()
+                except IntegrityError:
```
goosemania: Same as on line 322, I suggest to use `get_or_create` and create/update `deferred_artifact`.

pulp_file/app/models.py
```diff
@@ -248,3 +260,136 @@ def publish(self):
         Publish behavior for the file plugin has not yet been implemented.
         """
         raise NotImplementedError
+
+
+class BasicFileImporter(Importer, PulpManifestMixin):
+    """
+    Importer for "file" content.
```
goosemania: A docstring here is the same as for `FileImporter`, it will be helpful to have some details/specifics about this particular importer.

pulp_file/app/models.py
```diff
@@ -157,6 +122,53 @@ def _find_delta(manifest, inventory, mirror=True):
             removals = set()
         return Delta(additions=additions, removals=removals)
 
+
+
+class FileImporter(Importer, PulpManifestMixin):
+    """
+    Importer for "file" content.
```
goosemania: A docstring here is the same as for `BasicFileImporter`, it will be helpful to have some details/specifics about this particular importer.

pulp_file/app/models.py
```diff
@@ -248,3 +260,136 @@ def publish(self):
...
+                try:
+                    content_artifact.save()
+                except IntegrityError:
+                    pass
+                deferred_artifact.save()
+                # Report progress
+                progress_bar.increment()
+
+        # Remove content if there is any to remove
+        if delta.removals:
```
goosemania: This condition does not take into account importers `sync_mode` attribute. In [MVP doc](https://pulp.plan.io/projects/pulp/wiki/Pulp_3_Minimum_Viable_Product#Importers) `sync_mode` is additive by default, this importer works in mirror mode currently.

pulp_file/app/models.py
```diff
@@ -248,3 +260,136 @@ def publish(self):
...
+        """
+        inventory = self._fetch_inventory()
+        manifest = self._fetch_manifest()
+        delta = self._find_delta(manifest, inventory)
+        parsed_url = urlparse(self.feed_url)
+        root_dir = os.path.dirname(parsed_url.path)
+
+        # Start reporting progress
+        if self.download_policy != self.IMMEDIATE:
+            description = "Adding file content to the repository without downloading artifacts."
```
goosemania: We probably need to discuss how to deal with all the hard-coded strings, do we want them in a separate files or not. As well as how we are going to i18n them, `gettext`?

pulp_file/app/models.py
dkliban: I am adding more detailed docstrings.

pulp_file/app/models.py
dkliban: I'm adding a more detailed docstring.

pulp_file/app/models.py
dkliban: I will i18n using gettext, but I am not sure about storing the strings in a separate file.

pulp_file/app/models.py
```diff
@@ -248,3 +260,136 @@ def publish(self):
...
+            description = "Dowloading artifacts and adding content to the repository."
+        progress_bar = ProgressBar(message=description, total=len(delta.additions))
+        progress_bar.save()
+
+        # Add content and report progress while doing so
+        with progress_bar:
+            for entry in manifest.read():
+                key = Key(path=entry.path, digest=entry.digest)
+                if key not in delta.additions:
+                    continue
```
dkliban: i am glad you noticed this. i will add the size and digest to the 'additions' so we can iterate whole list just once.

pulp_file/app/models.py
dkliban: It needs to up here because it's needed for the deferred downloads. I added some clean up code.

pulp_file/app/models.py
dkliban: i updated both ContentArtifact and DeferredArtifact to use get_or_create.

pulp_file/app/models.py
dkliban: i added some cleanup code for unexpected errors on Artifact.save()

pulp_file/app/models.py
dkliban: done

pulp_file/app/models.py
dkliban: sync mode is not currently a property of an importer

pulp_file/app/models.py
```diff
@@ -17,11 +23,12 @@
 
 log = getLogger(__name__)
 
+BUFFER_SIZE = 65536
 
 # Changes needed.
 Delta = namedtuple('Delta', ('additions', 'removals'))
 # Natural key.
-Key = namedtuple('Key', ('path', 'digest'))
+FileTuple = namedtuple('Key', ('path', 'digest', 'size'))
```
jortel: The name `FileTuple` suggests a tuple of files.  I think `Key` was fine but if you want more clarity, how about `FileKey`?

dkliban: I wanted to add size attribute to it. I wasn't sure if the name Key was still appropriate after that. If you think it is, I can switch it back.

pulp_file/app/models.py
```diff
@@ -59,51 +66,10 @@ def natural_key(self):
         return Key(path=self.path, digest=self.digest)
 
 
-class FileImporter(Importer):
+class PulpManifestMixin(object):
```
jortel: Using mixins here seems overly complicated.  In general, multiple inheritance should be avoided given most programmer's understanding of method resolution order (MRO).  Something like this would be more straight forward and, I think a better fit:

```
    BaseImporter
     <abstract>
          |
    ---------------
    |             |
FileImporter  BasicFileImporter

```

It is my understanding that the purpose of writing the importer without using the changeset is:

- Enable earlier the merging of #3104 before the changeset updated for model changes.
- R&D for writing documentation.

I don't think there is value in maintaining 2 importers here long term.  Instead, I think it will confuse our plugin writer community.  I recommend:
```
class BasicFileImporter(FileImporter):

    def sync():
        # just override this.
```

and change as little else as possible since the `BasicFilePlugin` is only temporary.

pulp_file/app/models.py
jortel: Also, `size` is not part of the key for a File.

pulp_file/app/models.py
```diff
@@ -59,51 +66,10 @@ def natural_key(self):
...
     """
-    Importer for "file" content.
+    Provides functions for working with an importer's PULP_MANIFEST.
```
jortel: The name `PULP_MANIFEST` is just a naming convention.  Suggest just referring to the `manifest` in code.

pulp_file/app/models.py
```diff
@@ -248,3 +265,138 @@ def publish(self):
...
+            description = _("Dowloading artifacts and adding content to the repository.")
+        progress_bar = ProgressBar(message=description, total=len(delta.additions))
+        progress_bar.save()
+
+        # Add content and report progress while doing so
+        with progress_bar:
+            for entry in delta.additions:
+                path = os.path.join(root_dir, entry.path)
+                url = urlunparse(parsed_url._replace(path=path))
+                content, content_created = FileContent.objects.get_or_create(path=entry.path,
```
jortel: This should be fetched in batches to avoid thousands of separate queries.

pulp_file/app/models.py
```diff
@@ -248,3 +265,138 @@ def publish(self):
...
+        # Start reporting progress
+        if self.download_policy != self.IMMEDIATE:
+            description = _("Adding file content to the repository without downloading artifacts.")
+        else:
+            description = _("Dowloading artifacts and adding content to the repository.")
+        progress_bar = ProgressBar(message=description, total=len(delta.additions))
+        progress_bar.save()
+
+        # Add content and report progress while doing so
+        with progress_bar:
```
jortel: Suggest you use _transactions_ to avoid tens of thousands of commits.  Also, committing a content unit, it's artifacts and/or remote-artifacts will help ensure data integrity.

pulp_file/app/models.py
```diff
@@ -248,3 +265,138 @@ def publish(self):
...
+            for entry in delta.additions:
+                path = os.path.join(root_dir, entry.path)
+                url = urlunparse(parsed_url._replace(path=path))
+                content, content_created = FileContent.objects.get_or_create(path=entry.path,
+                                                                             digest=entry.digest)
+                # Add content to the repository
+                association = RepositoryContent(
```
jortel: What if a content unit is already in the repository?

pulp_file/app/models.py
```diff
@@ -248,3 +265,138 @@ def publish(self):
...
+                    repository=self.repository,
+                    content=content)
+                association.save()
+                # ContentArtifact is required for all download policies. It is used for publishing.
+                content_artifact, content_artifact_created = ContentArtifact.objects.get_or_create(
```
jortel: Content is immutable, this could blindly associate new artifacts to an existing content (unit).  This seems undesirable.

pulp_file/app/models.py
```diff
@@ -248,3 +265,138 @@ def publish(self):
...
+
+                try:
+                    artifact = Artifact.objects.get(sha256=entry.digest)
+                except Artifact.DoesNotExist:
+                    download = self.get_download(url, entry.path)
+                    download.validations.append(SizeValidation(entry.size))
+                    download.validations.append(DigestValidation('sha256', entry.digest))
+                    try:
+                        download()
+                    except DownloadError as e:
```
jortel: I think this approach of creating things then deleting them again on failure is subject to race conditions.  Especially without using transactions.

For example:

importer-1> creates content file-1.
importer-2> does get_or_create and finds file-1.
importer-1> fails to download an artifact and then deletes file-1.
importer-2> ties to add content-artifacts to a content that does not exist in the DB and fails.

Strongly recommend transactions instead.

pulp_file/app/models.py
jortel: The `size` is not part of the FileContent unique key.  Why do you want to add it here?

pulp_file/app/models.py
jortel: `get_or_create()` is subject to race conditions and can still raise `IntegretyError`.  This should be in a _try_ block and do a `get()` when `IntegretyError` is caught.

I would discourage using `get_or_create()` in concurrent systems.

https://docs.djangoproject.com/en/1.11/ref/models/querysets/#get-or-create

pulp_file/app/models.py
```diff
@@ -248,3 +265,138 @@ def publish(self):
...
+                association = RepositoryContent(
+                    repository=self.repository,
+                    content=content)
+                association.save()
+                # ContentArtifact is required for all download policies. It is used for publishing.
+                content_artifact, content_artifact_created = ContentArtifact.objects.get_or_create(
+                    content=content, relative_path=entry.path)
+                # DeferredArtifact is required for deferred download policies.
+                deferred_artifact, deferred_artifact_created = \
+                    DeferredArtifact.objects.get_or_create(
```
jortel: See previous comment re: using `get_or_create()`.

pulp_file/app/models.py
```diff
@@ -248,3 +265,138 @@ def publish(self):
...
+                    continue
+
+                try:
+                    artifact = Artifact.objects.get(sha256=entry.digest)
+                except Artifact.DoesNotExist:
+                    download = self.get_download(url, entry.path)
+                    download.validations.append(SizeValidation(entry.size))
+                    download.validations.append(DigestValidation('sha256', entry.digest))
+                    try:
+                        download()
```
jortel: Suggest concurrent downloading.

pulp_file/app/models.py
dkliban: We should enforce unique_together for the path and digest of FileContent. If we enforce it at database level, then we won't run into a problem.

pulp_file/app/models.py
dkliban: The content is not in the repository because we just checked in the code above. The tasking system makes sure that  only one task operates on a repo at a time. So we can be assured that this content is still not present.

pulp_file/app/models.py
jortel: There already is a unique constraint for (path, digest).

The problem is - content can be created by another process/thread between the get and create which will raise `IntegretyError` which is not caught here and the sync will fail.

pulp_file/app/models.py
jortel: True, but this assumes flawless logic in determining which content needs to be added to the repository.  I would still recommend try/except here for robustness but not required.

pulp_file/app/models.py
```diff
@@ -248,3 +265,138 @@ def publish(self):
...
+                    # Report progress
+                    progress_bar.increment()
+                    continue
+
+                try:
+                    artifact = Artifact.objects.get(sha256=entry.digest)
+                except Artifact.DoesNotExist:
+                    download = self.get_download(url, entry.path)
+                    download.validations.append(SizeValidation(entry.size))
+                    download.validations.append(DigestValidation('sha256', entry.digest))
```
jortel: These two validations will have already been setup by `Importer.get_download()`.  Don't need these 2 lines.

dkliban:
This is moving to pulp_example.

PR #10, opened 2017-08-14
Annotates and rearranges the FileImporter

mhrivnak:
There are no behavioral changes. This change:

- moves a small number of minor behaviors into pulpcore
- rearranges some code for clarity and readability
- adds in-line documentation and expands doc blocks to make this importer a good example

This depends on https://github.com/pulp/pulp/pull/3122

### DIFF
--- a/pulp_file/app/models.py
+++ b/pulp_file/app/models.py
@@ -1,7 +1,6 @@
 import os
 
 from collections import namedtuple
-from gettext import gettext as _
 from logging import getLogger
 from urllib.parse import urlparse, urlunparse
 
@@ -11,8 +10,6 @@ from pulpcore.plugin.models import Artifact, Content, Importer, Publisher
 from pulpcore.plugin.changeset import (
     BatchIterator,
     ChangeSet,
-    ChangeFailed,
-    ChangeReport,
     PendingArtifact,
     PendingContent,
     SizedIterable,
@@ -24,8 +21,6 @@ from pulp_file.manifest import Manifest
 log = getLogger(__name__)
 
 
-# Changes needed.
-Delta = namedtuple('Delta', ('additions', 'removals'))
 # Natural key.
 Key = namedtuple('Key', ('path', 'digest'))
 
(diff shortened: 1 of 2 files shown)


pulp_file/app/models.py
```diff
@@ -73,174 +68,152 @@ class FileImporter(Importer):
...
-                'r': removed,
-                'f': failed
-            })
-        # On failed > 0, raise a PulpCodedException?
-        # Done
+        FileSync(self).run()
 
-    def _fetch_inventory(self):
+
+class FileSync:
```
jortel: I think `Synchronizer` would be better.

pulp_file/app/models.py
```diff
@@ -73,174 +68,152 @@ class FileImporter(Importer):
...
+    """
+    Sync object for FileContent
```
jortel: Suggest: _Repository synchronizer for FileContent._  :)

pulp_file/manifest.py
```diff
@@ -79,12 +79,10 @@ def read(self):
...
-            for line in fp.readlines():
-                n += 1
+            for i, line in enumerate(fp.readlines(), 1):
                 line = line.strip()
                 if not line:
                     continue
                 if line.startswith('#'):
                     continue
-                yield Entry.parse(Line(number=n, content=line))
+                yield Entry.parse(Line(number=i, content=line))
```
jortel: Seems like _number=n_ is clearer.

mhrivnak: For some reason "i" and "j" are the convention, all the way back to its introduction: https://docs.python.org/2.3/whatsnew/section-enumerate.html I think that's a convention in algebra to use i and j for indicies. But any letter will do, and I'm happy going back to n. In particular, since this is a count (1-based) instead of an index (0-based), maybe n would help distinguish that.

pulp_file/app/models.py
mhrivnak: I'll give ``Synchronizer`` a try, and we can see how it looks.

mhrivnak:
@jortel what would you think if I moved the ``Synchronizer`` class into a separate module ``sync.py`` ? It's not a model, so I'm thinking that would make more sense than where it is right now. I think it's in ``models.py`` just from inertia, and it hadn't occurred to me to consider moving it.

mhrivnak:
@jortel here's what happened when I tried separating the Synchronizer to its own module. This is actually a bit of a problem for us I think.

I quickly found a circular import problem. The ``FileImporter`` wants to instantiate and use a ``Synchronizer``. The ``Synchronizer`` of course needs to use the data model (both the ``FileContent`` model itself and the corresponding ``Key`` namedtuple).

Thinking of the idea that circular imports are often a sign of poor design, I realized that there is a single and specific design choice throwing a wrench in the gears. We have a single class acting as both model and controller: the ``FileImporter``. It both defines part of the data model, AND it is expected to perform an action (sync) that could make use of several different models (classic controller territory).

We made the choice deliberately to put the plugin API methods (sync, upload, etc.) on the ``Importer`` to give the plugin writer a simple experience. But here we see that is already causing difficulty for plugin writing as well.

Perhaps we should re-consider having the ``Importer`` both act as model and controller. Any initial thoughts? My immediate impression is that this might make a good topic on pulp-dev@.

jortel:
Interesting.  Yes, let's discuss on pulp-dev.

PR #11, opened 2017-09-12
Support Importer.get_download() being temporarily removed.

jortel:
https://pulp.plan.io/issues/3007

### DIFF
--- a/pulp_file/app/models.py
+++ b/pulp_file/app/models.py
@@ -24,6 +24,8 @@ from pulpcore.plugin.changeset import (
     SizedIterable,
 )
 
+from pulpcore.plugin.download.futures import Factory
+
 from pulp_file.manifest import Manifest, Entry
 
 
@@ -81,6 +83,27 @@ class FileImporter(Importer):
         """
         Synchronizer(self).run()
 
+    def get_downloader(self, url, destination, artifact=None):
+        """
+        Get an appropriate download object based on the URL that is fully configured using
+        the importer attributes.  When an artifact is specified, the download is tailored
+        for the artifact.
+
+        Args:
+
+            url (str): The download URL.
+            destination (str): The absolute path to where the downloaded file is to be stored.
+            artifact (pulpcore.app.models.Artifact): An optional artifact.
+
+        Returns:
+            pulpcore.plugin.download.futures.Download: The appropriate download object.
+
+        Notes:
+            This method supports plugins downloading metadata and the
+            `streamer` downloading artifacts.
+        """
+        return Factory(self).build(url, destination, artifact)
+
 
 class Synchronizer:
     """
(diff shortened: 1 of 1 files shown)


pulp_file/app/models.py
```diff
@@ -137,8 +139,8 @@ def _fetch_manifest(self):
         Fetch (download) the manifest.
         """
         parsed_url = urlparse(self._importer.feed_url)
-        download = self._importer.get_download(
-            self._importer.feed_url, os.path.basename(parsed_url.path))
+        factory = Factory(self._importer)
```
bmbouter: Do you want to move the get_downloader implementation from core to the Importer you are defining instead of using the factory directly here? If we have an importer that has a get_downloader() method, either implementation would provide a get_downloader method that could eventually be used.

jortel: Ah, yeah, good point.  Yeah.  will do.

jortel:
@bmbouter, updated.

jortel:
The `Importer.get_download()` is used by the ChangeSet.  Let's discuss this.

bmbouter:
This should work even if we remove it from the main Importer in core. I think the changeset will only call get_downloader() after casting to the detail type which will cause it to use the importer you are defining. That importer has get_downloader() from this PR so this should work right?

jortel:
Yes, it will _work_ because the importer is already the concrete class.  But, I'm not convinced it's the best approach.  The `ChangeSet` works with the _futures_ download package and removing the `Importer.get_downloader()` all together, introduces an additional burden on every plugin writer.  They will need to implement get_downloader() to return a _futures_ downloader  in order to use the `ChangeSet`.  This seems undesirable.  Also, the method should be on the base class interface for the ChangeSet to use it and so that the plugin writer know what/how to override.  

I'm thinking we should leave it as is on the Importer (base) and just document that it can return either a _futures_ or _asyncio_ downloader for now.  We can also explain that this is temporary TBD on download API chosen.  This seems to be the easiest path forward.

bmbouter:
This isn't making sense to me. One get_downloader() implementation on the Importer in core can't provide both types of downloaders. The different downloaders use different factories so one get_downloader() implementation will return 1 type of downloader in all cases. This is why we needed to have two implementations of get_downloader() to start with. What do you think about this?

jortel:
Like this?
`Importer.get_futures_downloader()`
and
`Importer.get_asyncio_downloader()`

bmbouter:
Oh that sounds good. Would you want to make the PR against platform that introduces one or both of those? Also the change you made to pulp_file to add the get_downloader() should probably be removed and updated to use `Importer.get_futures_downloader()` yes? I can have my rebase add the `Importer.get_asyncio_downloader()`. How does this sound?

jortel:
Sounds good.

PR #27, opened 2018-01-16
Mark new versions complete after sync

asmacdo:
re #3222

### DIFF
--- a/pulp_file/app/tasks.py
+++ b/pulp_file/app/tasks.py
@@ -32,7 +32,7 @@ Key = namedtuple('Key', ('path', 'digest'))
 @shared_task(base=UserFacingTask)
 def sync(importer_pk):
     """
-    Validate the importer and create/update a RepositoryVersion.
+    Validate the importer, create and finalize RepositoryVersion.
 
     Args:
         importer_pk (str): The importer PK.
@@ -51,15 +51,14 @@ def sync(importer_pk):
 
     with transaction.atomic():
         new_version = models.RepositoryVersion(repository=importer.repository)
+        new_version.number = importer.repository.last_version + 1
+        importer.repository.last_version = new_version.number
         new_version.save()
-        # bump the repository's last_version
-        if importer.repository.last_version < new_version.number:
-            importer.repository.last_version = new_version.number
-            importer.repository.save()
+        importer.repository.save()
         created_resource = models.CreatedResource(content_object=new_version)
         created_resource.save()
-    synchronizer = Synchronizer(importer, new_version, base_version)
 
+    synchronizer = Synchronizer(importer, new_version, base_version)
     with working_dir_context():
         log.info(
             _('Starting sync: repository=%(repository)s importer=%(importer)s'),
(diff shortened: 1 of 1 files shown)


asmacdo:
Depends on https://github.com/pulp/pulp/pull/3268

pulp_file/app/tasks.py
```diff
@@ -32,10 +32,7 @@
 @shared_task(base=UserFacingTask)
 def sync(importer_pk):
     """
-    Call sync on the importer defined by a plugin.
-
-    Check that the importer has a feed_url, which is necessary to sync. A working directory
-    is prepared, the plugin's sync is called, and then working directory is removed.
+    Validate the importer and create/update a RepositoryVersion.
```
dkliban: I think this should just be create instead of create/update

pulp_file/app/tasks.py
```diff
@@ -48,15 +45,20 @@ def sync(importer_pk):
...
         new_version = models.RepositoryVersion(repository=importer.repository)
         new_version.save()
-        created = models.CreatedResource(content_object=new_version)
-        created.save()
-
-    base_version = None
-    with suppress(models.RepositoryVersion.DoesNotExist):
-        base_version = importer.repository.versions.latest()
+        # bump the repository's last_version
+        if importer.repository.last_version < new_version.number:
```
dkliban: What is the value of repository.last_version, the first time you are creating a version?
When would last_version not be less than new_version.number?

daviddavis: last_version defaults to 0 so if there are no versions, it should be 0. In theory, last_version should never be less than new_version.number since [number is last_version + 1](https://git.io/vNl5s). However, there may be cases where somehow this gets out of sync. I am a bit conflicted about this check.

dkliban: If this check fails, i think you have to stop everything. A new version that is just created should always be greater than the latest version.

pulp_file/app/tasks.py
```diff
@@ -48,16 +45,20 @@ def sync(importer_pk):
     if not importer.feed_url:
         raise ValueError(_("An importer must have a 'feed_url' attribute to sync."))
 
+    base_version = None
+    with suppress(models.RepositoryVersion.DoesNotExist):
+        base_version = importer.repository.versions.exclude(complete=False).latest()
```
daviddavis: I'm wondering if we should just call this 'old_version' to be consistent with what's in Synchronizer?

asmacdo: For pulpcore, we decided to use "base_version". I used that name since this portion of the code will be moved back to pulpcore. I am hoping that the Synchronizer will be updated to base_version to match pulpcore.

pulp_file/app/tasks.py
asmacdo: Rather than perform this check, I moved the RepositoryVersion.number setting logic out of RepositoryVersion.save().

pulp_file/app/tasks.py
daviddavis: Would it make sense to use base_version in Synchronizer then?

asmacdo: yes but I thought that would make more sense in another PR. I'm happy to make the change though, if you prefer.

daviddavis: Another PR is fine. Sorry to hold this up.

PR #28, opened 2018-01-22
3296 create publish task and route

asmacdo:
https://pulp.plan.io/issues/3296

depends on https://github.com/pulp/pulp/pull/3279


pulp_file/app/tasks.py
```diff
@@ -29,6 +32,100 @@
...
+
+    with transaction.atomic():
+        publication = models.Publication(publisher=publisher,
+                                         repository_version=repository_version)
+        publisher.publication = publication
+        publication.save()
+        created = models.CreatedResource(content_object=publication)
+        created.save()
+        with working_dir_context() as working_dir:
+            publisher.working_dir = working_dir
```
werwty: Why is this needed? I don't see working_dir on the publisher model, nor do I see this referenced anywhere else

pulp_file/app/tasks.py
```diff
@@ -29,6 +32,100 @@
...
+            manifest.write(_publish(publisher))
+            metadata = PublishedMetadata(
+                relative_path=os.path.basename(manifest.path),
+                publication=publisher.publication,
+                file=File(open(manifest.path, 'rb')))
+            metadata.save()
+            publisher.last_published = datetime.utcnow()
+            publisher.save()
+            distributions = models.Distribution.objects.filter(publisher=publisher)
+            distributions.update(publication=publication)
```
werwty: why update this automatically to the latest publication instead of letting the user create a new distribution if/when they want it?

pulp_file/app/tasks.py
asmacdo: I don't think it is necessary to set it on the publisher.

pulp_file/app/tasks.py
```diff
@@ -29,6 +32,100 @@
...
+        publication = models.Publication(publisher=publisher,
+                                         repository_version=repository_version)
+        publisher.publication = publication
+        publication.save()
+        created = models.CreatedResource(content_object=publication)
+        created.save()
+        with working_dir_context() as working_dir:
+            publisher.working_dir = working_dir
+            manifest = Manifest('PULP_MANIFEST')
+            manifest.write(_publish(publisher))
```
asmacdo: AFAIK this is the only part that really needs a working_dir

pulp_file/app/tasks.py
```diff
@@ -29,6 +32,107 @@
...
+
+    log.info(
+        _('Publishing: repository=%(repository)s, version=%(version)d, publisher=%(publisher)s'),
+        {
+            'repository': repository.name,
+            'publisher': publisher.name,
+            'version': repository_version.number,
+        })
+
+    with transaction.atomic():
```
jortel: I know you based this on code in core but;
- I don't think we _should_ create the entire publication in one transaction.  Perhaps commit like every 50-100 records?
- Add progress reporting?

pulp_file/app/tasks.py
```diff
@@ -29,6 +32,107 @@
...
+        with working_dir_context():
+            try:
+                manifest = Manifest('PULP_MANIFEST')
+                manifest.write(_publish(publication, repository))
+                metadata = PublishedMetadata(
+                    relative_path=os.path.basename(manifest.path),
+                    publication=publication,
+                    file=File(open(manifest.path, 'rb')))
+                metadata.save()
+                publisher.last_published = datetime.utcnow()
```
jortel: IIRC, we agreed to get rid of the Publisher.last_published.  It is redundant to the created timestamp on the publication.  Let's not set it here.

pulp_file/app/tasks.py
```diff
@@ -29,6 +32,107 @@
...
+
+    Returns:
+        Artifact: When the artifact exists.
+        RemoteArtifact: When the artifact does not exist.
+    """
+    artifact = content_artifact.artifact
+    if not artifact:
+        artifact = RemoteArtifact.objects.get(
+            content_artifact=content_artifact,
+            importer__repository=repository)
```
jortel: If the relationship between importer and repository is removed, We should probably add `RemoteArtifact.repository `.  That way we know which importer created the record and for which repository.  This would become: `repository=repository`.

pulp_file/app/viewsets.py
```diff
@@ -43,3 +44,14 @@ class FilePublisherViewSet(viewsets.PublisherViewSet):
     endpoint_name = 'file'
     queryset = models.FilePublisher.objects.all()
     serializer_class = serializers.FilePublisherSerializer
+
+    @decorators.detail_route(methods=('post',))
+    def publish(self, request, pk):
+        publisher = self.get_object()
+        repository_pk = str(publisher.repository.pk)
```
jortel: The redmine task indicates that passing `repository_version` should be supported.  What about that?

pulp_file/app/tasks.py
asmacdo: +1. After a discussion with @jortel, we agree that removal will need to happen in a separate issue.

pulp_file/app/tasks.py
asmacdo: Sounds good to me. I'd like to merge as is, so I have created a new story for this work:
https://pulp.plan.io/issues/3315

pulp_file/app/tasks.py
asmacdo: This was left over from auto_distribute. For now, I've removed it.

PR #29, opened 2018-01-24
Fix download example in documentation.

kersommoura:
Update the example of download to reflect the current stage of
development of Pulp.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -197,7 +197,7 @@ Check status of a task
 
 ``$ http GET http://localhost:8000/api/v3/tasks/82e64412-47f8-4dd4-aa55-9de89a6c549b/``
 
-Download ``foo.tar.gz`` from Pulp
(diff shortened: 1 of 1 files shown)


jortel:
LGTM

dralley:
I don't object to using ```test.iso``` instead of ```foo.tar.gz``` but why would we not change all of the other steps to match?   If we're using the README as a quickstart guide, you cannot download ```test.iso``` without first repeating all of the above steps.

We should make it consistent.

kersommoura:
@dralley, this PR was an attempt to fix an example that did not work. And `test.iso` is part of [fixtures](https://repos.fedorapeople.org/pulp/pulp/demo_repos/test_file_repo/) that was used as feed to the created repo. Then you can download `test.iso` as described in the documentation.

PR #32, opened 2018-01-25
Working sync with repoversion context manager

werwty:
Needs to be run with https://github.com/pulp/pulp/pull/3285

re #3285
https://pulp.plan.io/issues/3285

### DIFF
--- a/pulp_file/app/tasks.py
+++ b/pulp_file/app/tasks.py
@@ -18,6 +18,7 @@ from pulpcore.plugin.changeset import (
     SizedIterable,
 )
 from pulpcore.plugin.tasking import UserFacingTask, WorkingDirectory
+from pulpcore.plugin.repository import RepositoryVersion
 
 from pulp_file.app import models as file_models
 from pulp_file.manifest import Entry, Manifest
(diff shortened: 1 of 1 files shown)


pulp_file/app/tasks.py
```diff
@@ -100,7 +101,7 @@ def publish(publisher_pk, repository_pk):
         })
 
     with transaction.atomic():
-        publication = models.Publication(publisher=publisher, repository_version=repository_version)
+        publication = models.Publication(publisher=publisher, repository_version=repository_version._model)
```
werwty: Calling `repository_version._model` until the publication context manager is created to handle this nicely.

jortel: The `RepositoryVersion._model` is _protected_ and should not be used outside the class heirarchy.  Since access is needed, you should probably add a `RepositoryVersion.model` _public_ property.

pulp_file/app/tasks.py
```diff
@@ -116,6 +117,7 @@ def publish(publisher_pk, repository_pk):
             except Exception as e:
                 publication.delete()
                 created_resource.delete()
+                raise
```
jortel: Good catch.

pulp_file/app/tasks.py
```diff
@@ -140,37 +142,19 @@ def sync(importer_pk):
...
 
-    with transaction.atomic():
-        new_version = models.RepositoryVersion(repository=importer.repository)
-        new_version.number = importer.repository.last_version + 1
-        importer.repository.last_version = new_version.number
-        new_version.save()
-        importer.repository.save()
-        created_resource = models.CreatedResource(content_object=new_version)
-        created_resource.save()
+    with RepositoryVersion.create(importer.repository) as new_version:
```
jortel: 👍

PR #35, opened 2018-02-01
Use publishing context manager.

jortel:
https://pulp.plan.io/issues/3295

### DIFF
--- a/pulp_file/app/tasks.py
+++ b/pulp_file/app/tasks.py
@@ -7,12 +7,10 @@ from urllib.parse import urlparse, urlunparse
 
 from celery import shared_task
 from django.core.files import File
-from django.db import transaction
 from django.db.models import Q
 
 from pulpcore.plugin.models import (
     Artifact,
-    CreatedResource,
     RepositoryVersion,
     Publication,
     PublishedArtifact,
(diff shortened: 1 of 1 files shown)


pulp_file/app/tasks.py
```diff
@@ -48,9 +46,16 @@ def _publish(publication):
...
+        _artifact = content_artifact.artifact
+        if not _artifact:
+            _artifact = RemoteArtifact.objects.get(
+                content_artifact=content_artifact,
+                importer__repository=publication.repository_version.repository)
+        return _artifact
     for content in publication.repository_version.content:
         for content_artifact in content.contentartifact_set.all():
-            artifact = _find_artifact(content_artifact, publication.repository_version.repository)
+            artifact = find_artifact()
```
asmacdo: I think it would be easier to read if we pass content_artifact, repository as arguments instead of depending on the scope/state.

pulp_file/app/tasks.py
```diff
@@ -101,28 +85,19 @@ def publish(publisher_pk, repository_pk):
...
-                metadata.save()
-            except Exception as e:
-                publication.delete()
-                created_resource.delete()
-                raise
+    with WorkingDirectory():
+        with Publication.create(repository_version, publisher) as publication:
+            manifest = Manifest('PULP_MANIFEST')
+            manifest.write(_publish(publication))
+            metadata = PublishedMetadata(
```
asmacdo: I had originally hoped this would be accomplished by `publication.add_metadata(manifest.path)`. I don't think its really necessary for the plugin writer to know about and use the PublishedMetadata object, but leaving like this is probably more appropriate since the publication is a django model.

pulp_file/app/tasks.py
```diff
@@ -48,9 +46,16 @@ def _publish(publication):
     Yields:
         Entry: The manifest entry.
     """
+    def find_artifact():
```
asmacdo: IMO, this function doesn't really need a docstring for the file plugin, but it would be helpful for plugin writers who are using the file plugin as an example. If the plugin writer isn't already totally familiar with deferred downloading, this step could be surprising. (Especially because deferred downloading isn't well documented in Pulp 3 yet.)

PR #36, opened 2018-02-07
Removing association between publishers/importers and repos

daviddavis:
refs #3341
https://pulp.plan.io/issues/3341

Requires https://github.com/pulp/pulp/pull/3302

### DIFF
--- a/pulp_file/app/tasks.py
+++ b/pulp_file/app/tasks.py
@@ -49,9 +49,7 @@ def _publish(publication):
     def find_artifact():
         _artifact = content_artifact.artifact
         if not _artifact:
-            _artifact = RemoteArtifact.objects.get(
-                content_artifact=content_artifact,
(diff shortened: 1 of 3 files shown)


pulp_file/app/tasks.py
```diff
@@ -49,9 +49,7 @@ def _publish(publication):
     def find_artifact():
         _artifact = content_artifact.artifact
         if not _artifact:
-            _artifact = RemoteArtifact.objects.get(
-                content_artifact=content_artifact,
-                importer__repository=publication.repository_version.repository)
+            _artifact = RemoteArtifact.objects.filter(content_artifact=content_artifact).first()
```
daviddavis: Multiple RemoteArtifacts might be returned here so we need to use `first()` instead of `get()`.

jortel: 👍

PR #41, opened 2018-02-14
Refactored tasks.

jortel:
Taking a procedural approach to the publishing and syncing tasks in attempt to simplify the plugin.


pulp_file/app/tasks/synchronizing.py
```diff
@@ -0,0 +1,193 @@
...
+def build_additions(importer, manifest, delta):
+    """
+    Build the content to be added.
+
+    Args:
+        importer (FileImporter): An importer.
+        manifest (Manifest): The downloaded manifest.
+        delta (Delta): The set of Key to be added and removed.
+
+    Yields:
```
daviddavis: Should this be 'Returns'? I know that it returns an iterable though.

pulp_file/app/tasks/synchronizing.py
```diff
@@ -0,0 +1,193 @@
...
+def build_removals(base_version, delta):
+    """
+    Build the content to be removed.
+
+    Args:
+        base_version (RepositoryVersion):  The base repository version.
+        delta (Delta): The set of Key to be added and removed.
+
+    Yields:
+        SizedIterable: The FileContent to be removed from the repository.
```
daviddavis: Here too.

daviddavis:
Code looks good. Will test it out.

pulp_file/app/tasks/synchronizing.py
jortel: Yes.  Used to be a generator then changed it to returning the `SizedIterator` and missed the docstring.

pulp_file/app/tasks/synchronizing.py
jortel: Yes.  Used to be a generator then changed it to returning the `SizedIterator` and missed the docstring.

PR #47, opened 2018-02-26
Fix sync/publish REST schema.

jortel:
https://pulp.plan.io/issues/3351

### DIFF
--- a/pulp_file/app/viewsets.py
+++ b/pulp_file/app/viewsets.py
@@ -1,8 +1,11 @@
 from gettext import gettext as _
 
 from django_filters.rest_framework import filterset
+from rest_framework.compat import coreapi, coreschema
 from rest_framework.decorators import detail_route
-from rest_framework.exceptions import ValidationError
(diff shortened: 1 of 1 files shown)


pulp_file/app/viewsets.py
```diff
@@ -37,12 +40,25 @@ class FileImporterViewSet(ImporterViewSet):
...
+                        description=_('URI for the repository to be synchronized.')),
+                    required=True)
+            ]))
     def sync(self, request, pk):
         importer = self.get_object()
-        repository = self.get_resource(request.data['repository'], Repository)
+        try:
+            repository_uri = request.data['repository']
+        except KeyError:
+            raise ValidationError(detail=_('Repository URI must be specified.'))
```
daviddavis: @dralley pointed out to me that ValidationError should be namespaced under serializers. See https://git.io/vAKBA.

jortel: @daviddavis, Fixed.

PR #51, opened 2018-03-05
Limit FileContent api fields to relative_path and artifact

daviddavis:
This solves a number of issues such as differences between digest on FileContent vs Artifact.sha256, and FileContent.relative_path vs ContentArtifact.relative_path.

This also renames FileContent.path to FileContent.relative_path.

fixes #3428, #3410
https://pulp.plan.io/issues/3428
https://pulp.plan.io/issues/3410

### DIFF
--- a/pulp_file/app/serializers.py
+++ b/pulp_file/app/serializers.py
@@ -1,16 +1,23 @@
 from rest_framework import serializers
 
+from pulpcore.plugin.models import Artifact
 from pulpcore.plugin.serializers import ContentSerializer, ImporterSerializer, PublisherSerializer
 
 from .models import FileContent, FileImporter, FilePublisher
 
 
 class FileContentSerializer(ContentSerializer):
-    path = serializers.CharField()
-    digest = serializers.CharField()
+    relative_path = serializers.CharField(
+        help_text="Relative location of the file within the repository"
+    )
+    artifact = serializers.HyperlinkedRelatedField(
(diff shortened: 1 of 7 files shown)


pulp_file/app/serializers.py
```diff
@@ -6,11 +6,16 @@
...
+        help_text="Relative location of the file within the repository"
+    )
+    artifact = serializers.HyperlinkedIdentityField(
+        view_name='artifacts-detail',
+        help_text="Artifact file representing the physical content"
+    )
 
     class Meta:
-        fields = ContentSerializer.Meta.fields + ('path', 'digest')
+        fields = ('_href', 'notes', 'relative_path', 'artifact')
```
jortel: `notes` ?

daviddavis: It's one of the fields on ContentSerializer:

https://github.com/pulp/pulp/blob/3.0-dev/pulpcore/pulpcore/app/serializers/content.py#L14-L30

I have no idea what it's for? Should I remove it?

jortel: No idea :)

daviddavis: Checked with @dkliban and he said this is a holdover from pulp 2. Users are apparently using it to store metadata on content units.

jortel: Okay.  Let's remove the `notes` then.

PR #53, opened 2018-03-07
Problem: only the latest repository version can be published

dkliban:
Solution: add ability to publish any repository version

This patch adds an optional repository_version parameter for the 'publish' endpoint of the 'file' publisher.
The REST API accepts either 'repository' or 'repository_version', but not both at the same time.

closes #3324
https://pulp.plan.io/issues/3324

### DIFF
--- a/pulp_file/app/viewsets.py
+++ b/pulp_file/app/viewsets.py
@@ -6,7 +6,8 @@ from rest_framework.decorators import detail_route
 from rest_framework import serializers, status
 from rest_framework.response import Response
 
-from pulpcore.plugin.models import Artifact, Repository
+from pulpcore.plugin.models import Artifact, Repository, RepositoryVersion
+
 from pulpcore.plugin.viewsets import (
     ContentViewSet,
     ImporterViewSet,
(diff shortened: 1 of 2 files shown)


pulp_file/app/viewsets.py
```diff
@@ -6,7 +6,9 @@
 from rest_framework import serializers, status
 from rest_framework.response import Response
 
-from pulpcore.plugin.models import Artifact, Repository
+from pulpcore.plugin.models import Artifact, Repository, RepositoryVersion
+from rest_framework.exceptions import ValidationError
```
daviddavis: ValidationError ought to be namespaced on serializers:

https://github.com/encode/django-rest-framework/blob/master/rest_framework/exceptions.py#L131

There are some examples in this file.

pulp_file/app/viewsets.py
```diff
@@ -83,17 +85,35 @@ class FilePublisherViewSet(PublisherViewSet):
...
+        if 'repository_version' in request.data and request.data['repository_version']:
+            repository_version = self.get_resource(request.data['repository_version'],
+                                                   RepositoryVersion)
+
+        if repository and repository_version:
+            raise ValidationError("Either the 'repository' or 'repository_version' can be "
+                                  "specified - not both.")
+
+        if not repository:
+            repository = repository_version.repository
```
daviddavis: Drop this and call `repository_version.repository` on line 113.

daviddavis:
Thanks @dkliban!

PR #60, opened 2018-03-22
Define schema for detail routes.

dparalen:
closes #3351

### DIFF
--- a/pulp_file/app/viewsets.py
+++ b/pulp_file/app/viewsets.py
@@ -28,6 +28,39 @@ class FileContentFilter(filterset.FilterSet):
         ]
 
 
+class _RepositorySyncURLSerializer(serializers.Serializer):
+    repository = serializers.URLField(
+        help_text=_('A URI of the repository to be synchronized.'),
+        label=_('Repository'),
+        required=True,
+        error_messages={
+            'required': _('The repository URI must be specified.')
+        })
+
+
+class _RepositoryPublishURLSerializer(serializers.Serializer):
+    repository = serializers.URLField(
+        help_text=_('A URI of the repository to be published.'),
+        label=_('Repository'),
+        required=False
+    )
+    repository_version = serializers.URLField(
+        help_text=_('A URI of the repository version to be published.'),
+        label=_('Repository version'),
+        required=False
+    )
+
+    def validate(self, data):
+        repository = data.get('repository')
+        repository_version = data.get('repository_version')
+        if (repository and not repository_version) or (not repository and repository_version):
+            return data
+        raise serializers.ValidationError(
+            _("Either the 'repository' or 'repository_version' need to be specified "
+              "but not both.")
+        )
+
+
 class FileContentViewSet(ContentViewSet):
     endpoint_name = 'file'
     queryset = FileContent.objects.all()
@@ -57,16 +90,18 @@ class FileImporterViewSet(ImporterViewSet):
     queryset = FileImporter.objects.all()
     serializer_class = FileImporterSerializer
 
-    @detail_route(methods=('post',))
+    @detail_route(methods=('post',), serializer_class=_RepositorySyncURLSerializer)
     def sync(self, request, pk):
+        """
+        Synchronizes a repository. The ``repository`` field has to be provided.
+        """
         importer = self.get_object()
-        try:
-            repository_uri = request.data['repository']
-        except KeyError:
-            raise serializers.ValidationError(detail=_('Repository URI must be specified.'))
-        repository = self.get_resource(repository_uri, Repository)
+        serializer = _RepositorySyncURLSerializer(data=request.data)
+        serializer.is_valid(raise_exception=True)
+        repository_uri = serializer.data['repository']
         if not importer.feed_url:
             raise serializers.ValidationError(detail=_('A feed_url must be specified.'))
+        repository = self.get_resource(repository_uri, Repository)
         result = tasks.synchronize.apply_async_with_reservation(
             [repository, importer],
             kwargs={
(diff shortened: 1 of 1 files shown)


pulp_file/app/viewsets.py
```diff
@@ -57,16 +88,16 @@ class FileImporterViewSet(ImporterViewSet):
...
+    @detail_route(methods=('post',), serializer_class=_RepositorySyncURLSerializer)
     def sync(self, request, pk):
         importer = self.get_object()
-        try:
-            repository_uri = request.data['repository']
-        except KeyError:
-            raise serializers.ValidationError(detail=_('Repository URI must be specified.'))
-        repository = self.get_resource(repository_uri, Repository)
+        serializer = _RepositorySyncURLSerializer(data=request.data)
+        if not serializer.is_valid():
```
werwty: Probably just call `serializer.is_valid(raise_exception=True)` here

pulp_file/app/viewsets.py
```diff
@@ -28,6 +28,37 @@ class Meta:
         ]
 
 
+class _URLField(serializers.URLField):
```
werwty: Why subclass URLField instead of using default URLField?

pulp_file/app/viewsets.py
```diff
@@ -82,21 +113,24 @@ class FilePublisherViewSet(PublisherViewSet):
     queryset = FilePublisher.objects.all()
     serializer_class = FilePublisherSerializer
 
-    @detail_route(methods=('post',))
+    @detail_route(methods=('post',), serializer_class=_RepositoryPublishURLSerializer)
     def publish(self, request, pk):
+        serializer = _RepositoryPublishURLSerializer(data=request.data)
+        if not serializer.is_valid():
```
werwty: Probably just call serializer.is_valid(raise_exception=True) here

pulp_file/app/viewsets.py
```diff
@@ -82,21 +113,24 @@ class FilePublisherViewSet(PublisherViewSet):
...
+            raise serializers.ValidationError(detail=serializer.errors)
+        repository_uri = serializer.data.get('repository')
+        repository_version_uri = serializer.data.get('repository_version')
         publisher = self.get_object()
-        repository = None
-        repository_version = None
-        if 'repository' not in request.data and 'repository_version' not in request.data:
```
werwty: I would recommend cleaning up this logic a bit:
```
--- a/pulp_file/app/viewsets.py
+++ b/pulp_file/app/viewsets.py
@@ -121,23 +121,20 @@ class FilePublisherViewSet(PublisherViewSet):
         repository_uri = serializer.data.get('repository')
         repository_version_uri = serializer.data.get('repository_version')
         publisher = self.get_object()
-        repository, repository_version = None, None
+
         if not repository_uri and not repository_version_uri:
             raise serializers.ValidationError(_("Either the 'repository' or 'repository_version' "
                                               "need to be specified."))
 
-        if repository_uri:
-            repository = self.get_resource(repository_uri, Repository)
-
-        if repository_version_uri:
-            repository_version = self.get_resource(repository_version_uri, RepositoryVersion)
-
-        if repository and repository_version:
+        elif repository_uri and repository_version_uri:
             raise serializers.ValidationError(_("Either the 'repository' or 'repository_version' "
                                               "can be specified - not both."))
-
-        if not repository_version:
+        elif repository_uri:
+            repository = self.get_resource(repository_uri, Repository)
             repository_version = RepositoryVersion.latest(repository)
+        else:
+            repository_version = self.get_resource(repository_version_uri, RepositoryVersion)
+
```

dparalen:
@werwty thanks for the review and the kind words! :D
w/r the logic refactoring, I moved part of it to a custom validator which, [according to the docs](http://www.django-rest-framework.org/api-guide/serializers/#object-level-validation) should be the preferred way to do it; it gives me this nicely structured and (hopefully) self-explanatory no-no when I try to cheat:

```
(pulp) [vagrant@pulp3 pulpcore]$ phttp  :8000/api/v3/publishers/file/540f8af9-150a-4b1c-921a-da465248dc25/publish/ foo=bar
HTTP/1.0 400 Bad Request
Allow: POST, OPTIONS
Content-Length: 107
Content-Type: application/json
Date: Fri, 23 Mar 2018 12:25:56 GMT
Server: WSGIServer/0.2 CPython/3.6.4
Vary: Accept, Cookie
X-Frame-Options: SAMEORIGIN

{
    "non_field_errors": [
        "Either the 'repository' or 'repository_version' need to be specified but not both."
    ]
}
```

Unfortunately, I wasn't able to reproduce your error but maybe I'm running out-of-date core...
Please, let me know wdyt.

Cheers,
milan

pulp_file/app/viewsets.py
```diff
@@ -28,6 +28,39 @@ class Meta:
...
+        label=_('Repository'),
+        required=False
+    )
+    repository_version = serializers.URLField(
+        help_text=_('A URI of the repository version to be published.'),
+        label=_('Repository version'),
+        required=False
+    )
+
+    def validate(self, data):
```
werwty: :+1:

pulp_file/app/viewsets.py
```diff
@@ -82,27 +114,18 @@ class FilePublisherViewSet(PublisherViewSet):
     queryset = FilePublisher.objects.all()
     serializer_class = FilePublisherSerializer
 
-    @detail_route(methods=('post',))
+    @detail_route(methods=('post',), serializer_class=_RepositoryPublishURLSerializer)
     def publish(self, request, pk):
         publisher = self.get_object()
```
werwty: I would recommend adding a docstring here. This would show up in the auto docs like so:
screenshot from 2018-03-23 13-38-28

werwty:
@dparalen I'm still running into an error when publishing repository_versions (but not repository). My pulp repo is up to date, and pulp_file is checked out from this PR. Can you check to see if you can reproduce it?

dparalen:
@werwty I just wanted to let you know that indeed that's the issue you're running into and I've been running my core with the fix applied all the time 🙀
...updating

dparalen:
@werwty in case my update is OK, please merge this change as I don't have the commit  bit ;)

Thanks,
milan

werwty:
@dparalen ha! I also don't have the commit bit. @daviddavis Can you take care of this?

daviddavis:
👍

PR #61, opened 2018-03-23
implement master/detail tasks

asmacdo:


pulp_file/app/models.py
```diff
@@ -2,12 +2,18 @@
...
 
-from pulpcore.plugin.models import Content, ContentArtifact, Importer, Publisher
+from pulpcore.plugin.models import Content, ContentArtifact, Importer, Publisher, Repository
+from pulpcore.app.models.task import Task
 
 
 log = getLogger(__name__)
 
 
+class FileTask(Task):
```
asmacdo: This isn't a real type of task. It just allows the plugin writers to view v3/tasks/file/ that include all of them (sync, publish, etc). This type of task won't ever be created because the FileTaskViewSet does not have the create mixin.

pulp_file/app/serializers.py
```diff
@@ -1,9 +1,33 @@
...
+
+    importer = DetailRelatedField(
+        queryset=FileImporter.objects.all()
+    )
+    repository = serializers.HyperlinkedRelatedField(
+        view_name='repositories-detail',
+        queryset=Repository.objects.all(),
+    )
+
+    reservation_structure = ["repository", "importer"]
```
asmacdo: Used by pulpcore to create reserved tasks.

pulp_file/app/serializers.py
```diff
@@ -1,9 +1,33 @@
...
+    task_kwarg_structure = {'importer_pk': "importer.pk",
+                            'repository_pk': "repository.pk"}
+    celery_task = tasks.synchronize
```
asmacdo: This can deploy plugin defined tasks like it does here, or it could deploy core defined tasks. Add/Remove seems like a good candidate for a core defined task that would work for some simple plugins.

asmacdo:
Closing for now, this branch has gotten really out of date. I'll reopen if it becomes relevant for discussion again.

PR #62, opened 2018-04-02
Renames Importer to Remote

dkliban:
closes #3491
https://pulp.plan.io/issues/3491


jortel:
I realize this comment should be directed at the pulpcore but - Given the name changed to `Remote`, The `Remote.feed_url` attribute seems odd now.  I think `Remote.url` would be more appropriate.

dralley:
@jortel I agree.  Would you mind grooming the task?  https://pulp.plan.io/issues/3543

PR #64, opened 2018-04-08
Rename Remote.feed_url to Remote.url

dralley:
Requires PR: pulp/pulp#3423
Requires PR: PulpQE/pulp-smash#930

re: 3543
https://pulp.plan.io/issues/3543

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -80,7 +80,7 @@ Create a repository ``foo``
 Create a new remote ``bar``
 -----------------------------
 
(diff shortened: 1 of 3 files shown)


dralley:
I suppose "Requires PR" syntax does not work on all repositories yet?

dralley:
Failing because correct pulp-smash branch not being fetched, but the smash PR can be merged before everything else without issue.

PR #65, opened 2018-04-10
Remote url no longer needs to be validated before sync

daviddavis:
ref #3395
https://pulp.plan.io/issues/3395


dralley:
Travis is fixed, build should be restarted

PR #70, opened 2018-04-19
Use HyperlinkedRelatedField to validate relations

asmacdo:


daviddavis:
Any idea about the test failure?

PR #73, opened 2018-04-20
Don't run pulpcore smash tests for pulp-file

daviddavis:
Looking for feedback on this change. PRs shouldn't be touching pulpcore so running pulpcore tests against pulp_file seems useless.


daviddavis:
Thinking this change over some more, pulp_file could be inadvertently modifying pulpcore's behavior so maybe we should be running pulpcore tests against pulp_file?

dkliban:
@daviddavis you have a point there

daviddavis:
Thanks for the feedback @dkliban. Closing.

PR #80, opened 2018-04-24
Fix content API root

dralley:


dralley:
*sigh* one of the sqlite test runners timed out.

PR #83, opened 2018-04-26
Move artifact code into serializer

daviddavis:
ref #3607
https://pulp.plan.io/issues/3607

### DIFF
--- a/pulp_file/app/viewsets.py
+++ b/pulp_file/app/viewsets.py
@@ -3,10 +3,9 @@ from gettext import gettext as _
 from django.db import transaction
 from django_filters.rest_framework import filterset
 from rest_framework.decorators import detail_route
-from rest_framework import serializers, status
-from rest_framework.response import Response
+from rest_framework import serializers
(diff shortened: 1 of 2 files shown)


pulp_file/app/viewsets.py
```diff
@@ -79,21 +78,8 @@ class FileContentViewSet(ContentViewSet):
...
-        data = request.data
-        data['digest'] = artifact.sha256
-        serializer = self.get_serializer(data=data)
-        serializer.is_valid(raise_exception=True)
-        content = serializer.save()
-        content.artifact = artifact
-
-        headers = self.get_success_headers(request.data)
-        return Response(serializer.data, status=status.HTTP_201_CREATED, headers=headers)
+    def create(self, request, *args, **kwargs):
```
asmacdo: I think we could probably just remove this and let it be handled by the parent class. We should move the transaction to the serializer.

daviddavis:
This is stale

PR #85, opened 2018-04-26
Validates that repository has a version to publish

dralley:
Raise a ValidationError if you try to publish a repository which
does not have any versions.

closes #3583
https://pulp.plan.io/issues/3583


daviddavis:
Since this is validation, what about moving this into the serializer?

dralley:
We could do it this way.  Is there a way to, inside the serializer, just set the "repository_version" and the only use that?

dralley:
Is there a way to, inside the serializer, just set the "repository_version" and the only use that?  That would be ideal.

dralley:
@daviddavis please re-review

daviddavis:
Works great. Thanks @dralley.

PR #86, opened 2018-05-01
Updates to use PendingVerison

bmbouter:


daviddavis:
@bmbouter can we close this out?

bmbouter:
Yes this needs to be closed. Thanks for reminding.

PR #101, opened 2018-07-22
Update for django-filter 2.0

gmbnomis:
Adapt as done in PR pulp/pulp#3551 to make filters on content work again.


daviddavis:
Thank you @gmbnomis!

PR #102, opened 2018-07-23
Introducing asyncio stages

bmbouter:
These are the plugin changes that go with the core's use of DeclarativeVersion. See that work here: https://github.com/pulp/pulp/pull/3559

### DIFF
--- a/pulp_file/app/tasks/synchronizing.py
+++ b/pulp_file/app/tasks/synchronizing.py
@@ -42,6 +42,7 @@ class FileFirstStage(FirstStage):
 
     def __init__(self, remote):
         """
+        The first stage of a pulp_file sync pipeline.
 
         Args:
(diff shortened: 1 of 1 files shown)


pulp_file/app/tasks/synchronizing.py
```diff
@@ -44,148 +30,48 @@ def synchronize(remote_pk, repository_pk):
...
+
+    first_stage = FileFirstStage(remote)
+    DeclarativeVersion(first_stage, repository).create()
+
+
+class FileFirstStage(FirstStage):
+
+    def __init__(self, remote):
+        """
+
```
bmbouter: Needs docstring here.

bmbouter:
Thanks @dkliban for the review and testing!

PR #110, opened 2018-08-07
Adds mirror parameter to endpoint that deploys sync task

vdusek:
re #3494
https://pulp.plan.io/issues/3494

closes #3944
https://pulp.plan.io/issues/3944

### DIFF
--- a/pulp_file/app/tasks/synchronizing.py
+++ b/pulp_file/app/tasks/synchronizing.py
@@ -16,7 +16,7 @@ from pulp_file.manifest import Manifest
 log = logging.getLogger(__name__)
 
 
-def synchronize(remote_pk, repository_pk):
+def synchronize(remote_pk, repository_pk, mirror):
     """
     Sync content from the remote repository.
 
@@ -25,6 +25,7 @@ def synchronize(remote_pk, repository_pk):
     Args:
         remote_pk (str): The remote PK.
         repository_pk (str): The repository PK.
+        mirror (bool): True for mirror mode, False for additive.
 
     Raises:
         ValueError: If the remote does not specify a URL to sync.
@@ -37,7 +38,7 @@ def synchronize(remote_pk, repository_pk):
         raise ValueError(_('A remote must have a url specified to synchronize.'))
 
     first_stage = FileFirstStage(remote)
-    DeclarativeVersion(first_stage, repository).create()
+    DeclarativeVersion(first_stage, repository, mirror).create()
 
 
 class FileFirstStage(Stage):
--- a/pulp_file/app/viewsets.py
+++ b/pulp_file/app/viewsets.py
@@ -91,12 +91,14 @@ class FileRemoteViewSet(RemoteViewSet):
         serializer = RepositorySyncURLSerializer(data=request.data, context={'request': request})
         serializer.is_valid(raise_exception=True)
         repository = serializer.validated_data.get('repository')
+        mirror = serializer.validated_data.get('mirror', True)
         result = enqueue_with_reservation(
             tasks.synchronize,
             [repository, remote],
             kwargs={
                 'remote_pk': remote.pk,
-                'repository_pk': repository.pk
+                'repository_pk': repository.pk,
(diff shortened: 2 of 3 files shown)


pulp_file/app/tasks/synchronizing.py
```diff
@@ -16,14 +16,15 @@
...
-def synchronize(remote_pk, repository_pk):
+def synchronize(remote_pk, repository_pk, sync_mode):
     """
     Create a new version of the repository that is synchronized with the remote
     as specified by the remote.
 
     Args:
         remote_pk (str): The remote PK.
         repository_pk (str): The repository PK.
+        sync_mode (str): The synchronization mode, 'mirror' (default) or 'additive'.
```
daviddavis: I think additive is the default.

daviddavis:
I'm not seeing the sync_mode param in the api docs (`http :8000/pulp/api/v3/docs/api.json | grep sync_mode`). I think you may need to update the `RepositorySyncURLSerializer`.

Also, does the sync_mode parameter actually work? Not sure if this is in the scope of this PR—just curious.

pulp_file/app/tasks/synchronizing.py
vdusek: I tried following on the master (without ```sync_mode``` parameter). Sync this remote -  [file](https://repos.fedorapeople.org/pulp/pulp/fixtures/file/PULP_MANIFEST) and then this remote - [file-large](https://repos.fedorapeople.org/pulp/pulp/fixtures/file-large/PULP_MANIFEST) with one repo. This is the response for the repository versions:
```
...
"results": [
        {
            "_added_href": "http://localhost:8000/pulp/api/v3/repositories/1/versions/2/added_content/",
            "_content_href": "http://localhost:8000/pulp/api/v3/repositories/1/versions/2/content/",
            "_href": "http://localhost:8000/pulp/api/v3/repositories/1/versions/2/",
            "_removed_href": "http://localhost:8000/pulp/api/v3/repositories/1/versions/2/removed_content/",
            "content_summary": {
                "file": 10
            },
            "created": "2018-08-08T07:31:36.496343Z",
            "id": 2,
            "number": 2
        },
        {
            "_added_href": "http://localhost:8000/pulp/api/v3/repositories/1/versions/1/added_content/",
            "_content_href": "http://localhost:8000/pulp/api/v3/repositories/1/versions/1/content/",
            "_href": "http://localhost:8000/pulp/api/v3/repositories/1/versions/1/",
            "_removed_href": "http://localhost:8000/pulp/api/v3/repositories/1/versions/1/removed_content/",
            "content_summary": {
                "file": 3
            },
            "created": "2018-08-08T07:31:35.028869Z",
            "id": 1,
            "number": 1
        }
    ]
```
In additive mode I'd expect all 13 files in version 2, so it looks like a mirror for me, or I understand it wrong.

vdusek:
```RepositorySyncURLSerializer``` is updated, PR - [pulp/pull/3583](https://github.com/pulp/pulp/pull/3583). 

And yes, it works ([declarative_version.py#L110-L113](https://github.com/pulp/pulp/blob/master/plugin/pulpcore/plugin/stages/declarative_version.py#L110-L113)).

File: pulp_file/app/viewsets.py [hunk 2]
```diff
@@ -69,6 +69,7 @@ def sync(self, request, pk):
         """
         Synchronizes a repository. The ``repository`` field has to be provided.
         """
+        sync_mode = request.data.get('sync_mode', 'mirror').lower()
```
daviddavis: A couple things. First, it seems like the value is case sensitive so I think we can remove `lower()`? Also, would it be better to pull this field out of the serializer (ie `sync_mode = serializer.validated_data.get('sync_mode', 'mirror')`) instead of the request data?

daviddavis:
Tested and it works great 👍

pulp_file/app/viewsets.py
vdusek: Thanks, fixed.

PR #113, opened 2018-08-12
Update smash-tests w/ changes in plugin_template

dralley:


pulp_file/tests/functional/utils.py
```diff
@@ -48,6 +52,46 @@ def populate_pulp(cfg, url=None):
     return client.get(FILE_CONTENT_PATH)['results']
 
 
+def gen_file_remote(url=None, manifest_path='PULP_MANIFEST', **kwargs):
```
Ichimonji10: The `gen_*` functions are conceptually simple: each one will generate a sane dict of attributes for you, and then update that dict with args and kwargs.

With this in mind, I dislike how the `manifest_path` parameter is consumed in this function. The logic now goes like this: each one will generate a sane dict of attributes for you, and then update that dict with args and kwargs, except for certain functions which will use parameter A to munge parameter B and then discard parameter A.

This also makes it a bit harder to generate *invalid* dicts. Let's say that I want to generate the following fixture:

```python
{
    'url': 'https://repos.fedorapeople.org/pulp/pulp/fixtures/my-files/PULP_MANIFESTT',
    ...
}
```

Again, this dict intentionally has malformed attributes. That's useful for testing. If I execute this code:

```python
gen_file_remote(url='https://repos.fedorapeople.org/pulp/pulp/fixtures/my-files/PULP_MANIFESTT')
```

...then I expect to get the dict above. But instead, my desired URL is silently "corrected."

pulp_file/tests/functional/utils.py
```diff
@@ -48,6 +52,46 @@ def populate_pulp(cfg, url=None):
...
+    """Return a semi-random dict for use in creating a file Remote.
+
+    :param url: The URL of an external content source.
+    """
+    if url is None:
+        url = urljoin(FILE_FIXTURE_URL, manifest_path)
+    else:
+        url = urljoin(url, manifest_path)
+    remote = gen_remote(url)
+    file_extra_fields = {
```
Ichimonji10: What's the point in making `file_extra_fields`?

pulp_file/tests/functional/utils.py
dralley: I kind of agree and kind of disagree.

Pulp File kind of works a bit differently from the other plugins, in the respect that it wants you to point it directly at the manifest instead of looking for metadata files at a hardcoded path.  So, maybe an option would be to define a ```FILE_FIXTURE_MANIFEST_URL``` as the default (this would still avoid 95% of the scattered ```urljoin()``` calls) while not having the problem you mention.

Another solution: in #pulp-dev this morning we discussed changing the behavior of Pulp File itself, either by making it more like the other content types and looking for a hardcoded manifest name, or by making the manifest name a configurable option on the remote w/ a default, which would match what is done here.

pulp_file/tests/functional/utils.py
dralley: There isn't and it can be removed, at least as it currently stands.

Ichimonji10: I think this syntax is invalid:

```python
>>> def gen_dict(**kwargs):
...     my_dict = {**kwargs}
  File "<stdin>", line 2
    my_dict = {**kwargs}
                ^
SyntaxError: invalid syntax
```

pulp_file/tests/functional/utils.py
Ichimonji10: Let's go for the first solution. Make the `FILE_FIXTURE_MANIFEST_URL` point directly at `PULP_MANIFEST`. Please, and thank you. :smile:

pulp_file/tests/functional/utils.py
dralley: Which Python are you using?  I think it was added in Python 3.5 (which is the lowest version we're supporting in Pulp 3 due to async requirements).

Ichimonji10: Python 3.6.5

Ichimonji10: The syntax thing is an aside. What I'm getting at is that this code is wasteful. Try running this code:

```python
#!/usr/bin/env python
# coding=utf-8


def main(**kwargs):
    my_dict = {'one': 'one', 'two': 'two'}
    my_dict_updates = {**kwargs}
    my_dict.update(my_dict_updates)
    print(id(kwargs))
    print(id(my_dict))
    print(id(my_dict_updates))
    return my_dict


if __name__ == '__main__':
    print(main(two='twooo', three='three'))
```

You'll see that `kwargs`, `my_dict`, `my_dict_updates` all have separate IDs:

```console
$ ./test.py 
140592596980000
140592596980144
140592596980072
{'one': 'one', 'two': 'twooo', 'three': 'three'}
```

dralley: >It appears that the Python 3.6.5 interpreter on my Xubuntu system considers this to be invalid syntax, but the Python 3.6.6 and 3.7.0 interpreters on my Arch Linux system consider this to be valid syntax.

LOL, oh, joy...  I'm not sure what action to take there... bug report?  Avoid the syntax?  Ugh.

Anyway, it's wasteful but I was kind of ambivalent about removing it.  One because I wanted you to see what the template does just generally speaking, but two, in a suite with so many hits to the network and database, an extra copy of a small 250 byte dictionary isn't really that significant.

pulp_file/tests/functional/utils.py
dralley: Solution #1 implemented, Cap'n.

pulp_file/tests/functional/api/test_crud_remotes.py
```diff
@@ -137,7 +137,7 @@ def _gen_verbose_remote():
 
     Note that 'username' and 'password' are write-only attributes.
     """
-    attrs = gen_remote(choice((FILE_FIXTURE_URL, FILE2_FIXTURE_URL)))
+    attrs = gen_file_remote(url=choice((FILE_FIXTURE_MANIFEST_URL, FILE2_FIXTURE_MANIFEST_URL)))
```
dralley: Thoughts on dropping the ```_URL```?

pulp_file/tests/functional/utils.py
dralley: Removed

pulp_file/tests/functional/api/test_crud_remotes.py
Ichimonji10: Pulp Smash typically appended `_URL` to fully-formed valid URLs. But local consistency is more important than what some other project does. Your call.

dralley: I'll leave it

pulp_file/tests/functional/utils.py
Ichimonji10: >  One because I wanted you to see what the template does just generally speaking

And now I know!

>  in a suite with so many hits to the network and database, an extra copy of a small 250 byte dictionary isn't really that significant.

True. I think that I come up with suggestions like this mainly because [thinking about memory](https://people.freebsd.org/~lstewart/articles/cpumemory.pdf) is a habit of mine. I also must point out that code can be consumed in ways other than "the Python interpreter executed it." For example, linters also consume code, and they anecdotally run faster when presented with efficient code.

dralley: Good point!

pulp_file/tests/functional/api/test_sync.py
```diff
@@ -33,119 +30,44 @@ def setUpClass(cls):
...
+        self.assertEqual(len(get_added_content(repo)), FILE_FIXTURE_COUNT)
 
         # Sync the repository again.
         latest_version_href = repo['_latest_version_href']
         sync(self.cfg, remote, repo)
         repo = client.get(repo['_href'])
-        self.assertNotEqual(latest_version_href, repo['_latest_version_href'])
-
-
-class SyncChangeRepoVersionTestCase(unittest.TestCase):
```
dralley: These two test cases don't test anything plugin-specific, so their results should be plugin-independent.  Therefore it would make sense to move them into pulpcore's "using_plugin" test suite.

I've moved the MultiResourceLocking test into a new file called using_plugin/test_tasking.py, and I moved this test into using_plugin/test_repo_versions.py.

PR #115, opened 2018-09-06
Allow custom name of the manifest when syncing and publishing

vdusek:
When creating a remote user can specify his own name of the manifest.
As a path to it will be used url/manifest. For example:
```
$ http POST .../remotes/file/ 
    name=rem1 
    url=https://repos.fedorapeople.org/pulp/pulp/demo_repos/test_file_repo/
    manifest=PULP_MANIFEST
```

Same situation when creating a publisher.
For example:
```
$ http POST .../publishers/file/ 
    name=publisher1 
    manifest=my_manifest.csv
```

closes #3912, #3913
https://pulp.plan.io/issues/3912
https://pulp.plan.io/issues/3913

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -93,7 +93,7 @@ Create a repository ``foo``
 Create a new remote ``bar``
 ---------------------------
 
-``$ http POST http://localhost:8000/pulp/api/v3/remotes/file/ name='bar' url='https://repos.fedorapeople.org/pulp/pulp/demo_repos/test_file_repo/PULP_MANIFEST'``
+``$ http POST http://localhost:8000/pulp/api/v3/remotes/file/ name='bar' url='https://repos.fedorapeople.org/pulp/pulp/demo_repos/test_file_repo/' manifest=PULP_MANIFEST``
 
 .. code:: json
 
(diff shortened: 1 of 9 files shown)


vdusek:
There's a problem in entering manifest param to the publish request. Example:
```
$ http POST $PUBLISHER_HREF'publish/' repository=$REPO_HREF manifest='my_manifest.csv'

...
{
    "manifest": [
        "Unexpected field"
    ]
}
```
The cause of the problem is here:
https://github.com/pulp/pulp_file/blob/3fdb37c62b46f4420dc7542807ac8c3135cabf19/pulp_file/app/viewsets.py#L133
Manifest doesn't pass through serializer validation. It's not part of the model. Is there another way how to get data from command line in the viewset?

vdusek:
Update: I didn't notice that there is a suggested solution in [issues/3913](https://pulp.plan.io/issues/3913). So name of the manifest is passed as a field of the publisher.

dkliban:
Storing a field on the FilePublisher seems better to me.

jortel:
:+1: adding a `manifest` field to `FilePublisher`.

vdusek:
Cool, thanks, so I'll update the tests and `README.rst`.

PR #116, opened 2018-09-11
Add minor changes to the functional tests

kersommoura:
Sort importers, reduce line length, and add assertions.


kersommoura:
Tested using source installation
``` python
python -m unittest discover pulp_file/tests/functional/
```

PR #123, opened 2018-09-18
Add test to sync an invalid feed iso

kersommoura:
Add a test to perform `sync` on an invalid feed iso. Refactor the
`SyncInvalidURLTestCase` to fit the new test.

Make assertions about the error message.

'[noissue]'

### DIFF
--- a/pulp_file/tests/functional/api/test_crud_remotes.py
+++ b/pulp_file/tests/functional/api/test_crud_remotes.py
@@ -137,7 +137,10 @@ def _gen_verbose_remote():
 
     Note that 'username' and 'password' are write-only attributes.
     """
-    attrs = gen_file_remote(url=choice((FILE_FIXTURE_MANIFEST_URL, FILE2_FIXTURE_MANIFEST_URL)))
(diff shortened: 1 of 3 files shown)


kersommoura:
Ping @qe

kersommoura:
Tested against source installation F28.

kersommoura:
@daviddavis, can you review/merge this pr?

dralley:
@kersommoura Do you think it's worth leaving a distinction between an invalid remote (like a URL that doesn't exist, invalid username/password credentials, etc.), and an invalid repository (files in the manifest that don't exist, have checksums that don't match, etc.)?

kersommoura:
@dalley, I agree. That is why I kept a different test for each o them.  They just are part of the same class - Since they are related to `sync` and produce the same kind of failure. Perhaps, I can update the names, and doc strings to make this clear.

dralley:
I meant by having two different test classes, one for different types of invalid repositories, and the other for different types of invalid remote configurations.

PR #130, opened 2018-10-12
Remove blank argument from model

pavelpicka:
as 'blank=True' is used for validating by
django form (include django admin) we do not
need to use it as checking by djangorestframewrok

https://pulp.plan.io/issues/4041

re: #4041

Signed-off-by: Pavel Picka <ppicka@redhat.com>


daviddavis:
What about these fields?

https://github.com/pulp/pulp_file/blob/76cc979e67fde128f78f3274697a4ea78f2269ec/pulp_file/app/models.py#L25-L26

PR #131, opened 2018-11-01
Remove 'null' argument from manifest model

pavelpicka:
manifest cannot be ever null as is needed
for file list

https://pulp.plan.io/issues/4079

closes #4079

Signed-off-by: Pavel Picka <ppicka@redhat.com>


daviddavis:
@pavelpicka let me know if I should merge this. I think it will create a conflict though with #130.

PR #132, opened 2018-11-14
Adds lazy download support

bmbouter:
Required PR: https://github.com/pulp/pulp/pull/3738

pulp_file now meaningfully uses the 'policy' attribute when syncing
content.

This is meant to be used with this PR:
https://github.com/pulp/pulp/pull/3738

https://pulp.plan.io/issues/4160
closes #4160

### DIFF
--- a/pulp_file/app/tasks/synchronizing.py
+++ b/pulp_file/app/tasks/synchronizing.py
@@ -4,7 +4,7 @@ import os
 from gettext import gettext as _
 from urllib.parse import urlparse, urlunparse
 
-from pulpcore.plugin.models import Artifact, ProgressBar, Repository
+from pulpcore.plugin.models import Artifact, ProgressBar, Remote, Repository
 from pulpcore.plugin.stages import (
     DeclarativeArtifact, DeclarativeContent, DeclarativeVersion, Stage
 )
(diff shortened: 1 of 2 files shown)


pulp_file/app/tasks/synchronizing.py
```diff
@@ -38,7 +38,11 @@ def synchronize(remote_pk, repository_pk, mirror):
         raise ValueError(_('A remote must have a url specified to synchronize.'))
 
     first_stage = FileFirstStage(remote)
-    DeclarativeVersion(first_stage, repository, mirror).create()
+    if remote.policy == 'immediate':
+        dv = DeclarativeVersion(first_stage, repository, mirror=mirror)
+    else:
+        dv = DeclarativeVersion(first_stage, repository, mirror=mirror, download_artifacts=False)
+    dv.create()
```
jortel: Suggest:
```
download_artifacts = (remote.policy == POLICY_MODE.IMMEDIATE)
dv = DeclarativeVersion(first_stage, repository, mirror=mirror, download_artifacts=download_artifacts)
```
to reduce code duplication.

Let's compare using the `constants.POLICY_MODE ` instead of the string literal.

bmbouter: Actually the Django docs show putting these types of things on the models themselves. After just importing a bunch of things "through" the plugin API, it seems easier just to do that instead. What about that?

jortel: Yeah, having the constants on a model would be good.

PR #141, opened 2018-12-10
Adding codecov and badges

daviddavis:
[noissue]


daviddavis:
Codecov seems to be only examining 3 files and it's reporting 100% test coverage. Need to figure out what's wrong.

PR #142, opened 2018-12-10
Fixing duplicate content exception

daviddavis:
fixes #3436
https://pulp.plan.io/issues/3436

### DIFF
--- a/pulp_file/app/serializers.py
+++ b/pulp_file/app/serializers.py
@@ -1,3 +1,5 @@
+from gettext import gettext as _
+
 from rest_framework import serializers
 
 from pulpcore.plugin.models import Artifact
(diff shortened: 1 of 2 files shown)


pulp_file/app/serializers.py
```diff
@@ -25,6 +27,19 @@ class FileContentSerializer(ContentSerializer):
         queryset=Artifact.objects.all()
     )
 
+    def validate(self, data):
+        """Validate the FileContent data"""
+        artifact = data['artifact']
+        content = FileContent.objects.filter(digest=artifact.sha256,
+                                             relative_path=data['relative_path']).first()
```
jortel: suggest `count()` would be more sufficient and more efficient.

jortel:
This approach will be subject to race conditions and only (but significantly) reduce the number of 500.  Perhaps catching the `IntegretyError` would be a better approach?

daviddavis:
@jortel fixed flake8 errors, addressed your feedback, and opened https://pulp.plan.io/issues/4264.

PR #148, opened 2018-12-14
Prepend Content model hierarchy with _

CodeHeeler:
Prepend all pulpcore model fields in the Content model hierarchy with _
and do the same for all plugins reliant on them.

re #4206
https://pulp.plan.io/issues/4206


pulp_file/app/serializers.py
```diff
@@ -41,7 +41,7 @@ def validate(self, data):
         return data
 
     class Meta:
-        fields = tuple(set(ContentSerializer.Meta.fields) - {'artifacts'}) + ('relative_path',
+        fields = tuple(set(ContentSerializer.Meta.fields) - {'_artifacts'}) + ('relative_path',
                                                                               'artifact')
```
dralley: Possibly 'artifact' should become '_artifact' also.  To be honest, I don't much like the way we replace this field name on some plugins, but this would at least be more consistent.

daviddavis: I'm a bit mixed about prefixing artifact with an underscore. I think it shouldn't be prefixed since it's defined here in the plugin but I guess prefixing it would make it more consistent from a user perspective

What would you suggest instead of creating an artifact field? Previously we were submitting a dictionary of artifacts when creating a content unit and it caused problems:

https://pulp.plan.io/issues/3425
https://pulp.plan.io/issues/3422
https://pulp.plan.io/issues/3423

CodeHeeler: Yeah, for 4206 we have not changed related names, only the target ones so far, but it does lead to some weird situations like this where you are unsure if something will need a _ without checking each field.  I'd be ok with this being an additional change later with discussion, but for now just trying to get 4206 as stated merged to unblock additional work.

dralley: @daviddavis The thing is that even though it's "from the plugin" it's replacing something that's from core.  So for some plugins you access the artifacts using "artifact" and for some you access the artifacts using "_artifacts"

If it causes problems then what is our solution to that problem for the content types that are intended to be associated with multiple artifacts?  Does it remain a problem, but one that we're sidestepping/kicking the can on by doing it this way for all of the current plugins?

daviddavis: > The thing is that even though it's "from the plugin" it's replacing something that's from core. So for some plugins you access the artifacts using "artifact" and for some you access the artifacts using "_artifacts"

Agreed, no objection from me.

> If it causes problems then what is our solution to that problem for the content types that are intended to be associated with multiple artifacts? Does it remain a problem, but one that we're sidestepping/kicking the can on by doing it this way for all of the current plugins?

Two of the problems were specific to the file plugin (in that the natural key involved relative_path and digest which also exist on artifact). The other problem was that files could be created with multiple artifacts. So those problems won't exist with multiple artifact content types.

daviddavis: I'm not totally happy with the `artifact` field solution myself but I couldn't find a better way to solve the problem. If you have a better suggestion, feel free to share it or open a bug. I think we could address it before the RC.

dralley: Ok, I'm fine with going forwards as-is.

PR #152, opened 2018-12-19
Expand duplicate content unit test case

kersommoura:
Expand newly added duplicate content unit test case. Add a test case to
confirm that content units using the same artifact, but different
relative_paths can be create without raise an HTTP exception.

https://pulp.plan.io/issues/4125
ref: #4125

### DIFF
--- a/pulp_file/tests/functional/api/test_crud_content_unit.py
+++ b/pulp_file/tests/functional/api/test_crud_content_unit.py
@@ -99,36 +99,45 @@ class ContentUnitTestCase(unittest.TestCase):
 
 
 class DuplicateContentUnit(unittest.TestCase):
-    """Attempt to create a duplicate content unit."""
+    """Attempt to create a duplicate content unit.
 
-    def test_all(self):
-        """Attempt to create a duplicate content unit.
+    This test targets the following issues:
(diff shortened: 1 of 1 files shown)


kersommoura:
ping @pulp/qe 

I added an extra test case to cover the negative case.

@daviddavis, please, can you review/merge this one?

dralley:
@kersommoura @daviddavis Should this test case be something we duplicate for each plugin, or would it be sufficient to have one copy of this and move it to pulpcore/tests/functional/using_plugin?

kersommoura:
@dralley , after finishing this test here. I intend to add to the: pulp_docker, pulp_rpm, and pulp_template. 
If there is a different understanding I can move the test as well.

daviddavis:
@dralley I think this test is going to be specific to each plugin. [The error message will be different](https://github.com/pulp/pulp_file/blob/master/pulp_file/app/serializers.py#L37-L40) and it sounds like some Content might not even have natural keys.

kersommoura:
I rebased this PR again.

@dralley, please, can you merge it?

daviddavis:
Done. Thanks :+1:

PR #153, opened 2019-01-03
Changing relative_path and digest to CharFields

daviddavis:
The relative_path and digest fields are text fields and are used in a constraint which MySQL doesn't
support.

### DIFF
--- a/pulp_file/app/models.py
+++ b/pulp_file/app/models.py
@@ -22,8 +22,8 @@ class FileContent(Content):
 
     TYPE = 'file'
 
-    relative_path = models.TextField(null=False)
-    digest = models.TextField(null=False)
+    relative_path = models.CharField(max_length=255, null=False)
+    digest = models.CharField(max_length=255, null=False)
 
     @property
     def artifact(self):


pulp_file/app/models.py
```diff
@@ -22,8 +22,8 @@ class FileContent(Content):
 
     TYPE = 'file'
 
-    relative_path = models.TextField(null=False)
-    digest = models.TextField(null=False)
+    relative_path = models.CharField(max_length=255, null=False)
+    digest = models.CharField(max_length=255, null=False)
```
dralley: If "digest" is expected to always be a sha256, we can just use 64, the length of a sha256 hash.

Also I'm sure 255 chars is more than relative_path will typically need but we may want to document that limitation, and maybe try to think if it could ever be violated.

daviddavis: I opened a PR for the digest change and an issue to discuss the relative_path problem.

PR: https://github.com/pulp/pulp_file/pull/191

Issue: https://pulp.plan.io/issues/4544

PR #155, opened 2019-01-15
Adapt to changes in Stages API

mdellweg:
Stages api hides the use of queues from users. Therefore custom stages
must be changed.

DO NOT MERGE YET

[noissue]


bmbouter:
The core commit was merged, so I think this needs to also be merged.

mdellweg:
I was about to remove the 'DO NOT MERGE YET'. But being on a conference i had other priorities. Sorry 'bout that.

PR #160, opened 2019-01-27
Remove policy from quickstart remote creation

asmacdo:
The streamed policy is currently broken, but also IMO "immediate" (being the default) should be the policy that we show for the quickstart.
[noissue]


bmbouter:
Removing this makes sense to me. The default will apply, and we can improve the lazy support docs for pulp_file specifically later.

PR #163, opened 2019-02-04
Bumps version for beta 8

dkliban:
[no issue]


daviddavis:
This LGTM. At some point we have to address https://pulp.plan.io/issues/4353 though.

daviddavis:
Travis is failing because you have `[no issue]` instead of `[noissue]`.

PR #165, opened 2019-02-06
Version bump

dralley:
[noissue]

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,7 @@
 from setuptools import setup, find_packages
 
 requirements = [
-    'pulpcore-plugin==0.1.0b18',
+    'pulpcore-plugin==0.1.0b19',
 ]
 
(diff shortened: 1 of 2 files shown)


dralley:
@dkliban This has to be version bumped too, since it depends on a package, that has a broken dependency with another package.

setup.py
```diff
@@ -3,15 +3,15 @@
 from setuptools import setup, find_packages
 
 requirements = [
-    'pulpcore-plugin==0.1.0b18',
+    'pulpcore-plugin>=0.1.0b19',
 ]
```
asmacdo: I think it's best to continue pinning at least until RC. Could you explain why?

dralley: https://travis-ci.org/pulp/ansible-pulp3/jobs/489343170#L5606

PR #166, opened 2019-02-06
Let ContentSerializer save the artifacts

mdellweg:
Required PR: pulp/pulpcore-plugin#58
Required PR: pulp/pulp_file#166

re #4394
https://pulp.plan.io/issues/4394

### DIFF
--- a/pulp_file/app/serializers.py
+++ b/pulp_file/app/serializers.py
@@ -5,7 +5,8 @@ from rest_framework import serializers
 from pulpcore.plugin.serializers import (
     SingleArtifactContentSerializer,
     RemoteSerializer,
-    PublisherSerializer
+    PublisherSerializer,
+    relative_path_validator,
 )
 
 from .models import FileContent, FileRemote, FilePublisher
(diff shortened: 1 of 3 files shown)


pulp_file/app/serializers.py
```diff
@@ -22,8 +23,13 @@ class FileContentSerializer(SingleArtifactContentSerializer):
 
     def validate(self, data):
         """Validate the FileContent data."""
-        artifact = data['_artifact']
-        content = FileContent.objects.filter(digest=artifact.sha256,
+        data = super().validate(data)
+
+        data['digest'] = data['_artifact'].sha256
+        data['_relative_path'] = data['relative_path']
+        relative_path_validator(data['_relative_path'])
```
jortel: The `SingleArtifactContentSerializer._artifact` field is already configured with the `relative_path_validator`.  Why does this needs to be called here?

mdellweg: The field validation happens before entering this `validate` method. And since we removed `_relative_path` in favor of `relative_path` from `Meta.fields`, it is not automatically validated. But now that you bring it up, we can just validate `relative_path` and we are safe.
Will upload that in a minute.

mdellweg: @jortel since that was not covered by test, i added one.

PR #169, opened 2019-02-13
Fix the build of `content_host` URL.

rochacbruno:
Fix the build of `content_host` URL using the distribution `base_path`

Required PR: https://github.com/PulpQE/pulp-smash/pull/1169

Relates to: https://github.com/PulpQE/pulp-smash/pull/1166

[noissue]

### DIFF
--- a/pulp_file/tests/functional/api/test_download_content.py
+++ b/pulp_file/tests/functional/api/test_download_content.py
@@ -8,6 +8,7 @@ from urllib.parse import urljoin
 from pulp_smash import api, config, utils
 from pulp_smash.pulp3.constants import DISTRIBUTION_PATH, REPO_PATH
 from pulp_smash.pulp3.utils import (
+    download_content_unit,
     gen_distribution,
     gen_repo,
     publish,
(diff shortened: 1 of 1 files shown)


pulp_file/tests/functional/api/test_download_content.py
```diff
@@ -92,9 +93,11 @@ def test_all(self):
         # …and Pulp.
         client.response_handler = api.safe_handler
 
-        unit_url = cfg.get_hosts('api')[0].roles['api']['scheme']
-        unit_url += '://' + distribution['base_url'] + '/'
-        unit_url = urljoin(unit_url, unit_path)
+        unit_url = reduce(urljoin, (
```
dralley: Do you think this code would be a good candidate for making into a utility function?  Such as

```
content = download_content_unit(distribution, unit_path)
pulp_hash = hashlib.sha256(content).hexdigest()
```
Although, I am fine with this as it is.  LGTM, up to you.

dralley: Or ```download_from_distribution()```, or ```get_distribution_url()``` and then download separately, etc.

kersommoura: I am fine with either solution.

rochacbruno: @dralley @kersommoura I can make it a function inside pulp_smash but the function will also need to take cfg as parameter

rochacbruno:
@kersommoura @dralley changes implemented

PR #172, opened 2019-02-14
Add test to publish a lazy synced repo

kersommoura:
Add tests to verify that a lazy synced repository can be published.

https://pulp.plan.io/issues/4418
ref: #4418

### DIFF
--- a/pulp_file/tests/functional/api/test_download_policies.py
+++ b/pulp_file/tests/functional/api/test_download_policies.py
@@ -8,40 +8,56 @@ from pulp_smash.pulp3.utils import (
     gen_repo,
     get_added_content_summary,
     get_content_summary,
+    publish,
     sync,
 )
 
 from pulp_file.tests.functional.constants import (
     FILE_FIXTURE_SUMMARY,
-    FILE_REMOTE_PATH
+    FILE_PUBLISHER_PATH,
+    FILE_REMOTE_PATH,
 )
-from pulp_file.tests.functional.utils import gen_file_remote
+from pulp_file.tests.functional.utils import (
+    gen_file_publisher,
+    gen_file_remote,
+)
 from pulp_file.tests.functional.utils import set_up_module as setUpModule  # noqa:F401
 
 
-class SyncDownloadPolicyTestCase(unittest.TestCase):
-    """Sync a repository with different download policies.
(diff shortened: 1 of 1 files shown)


pulp_file/tests/functional/api/test_download_policies.py
```diff
@@ -86,3 +88,44 @@ def do_test(self, download_policy):
...
+    def test_streamed(self):
+        """Publish a repo synced using ``streamed``.  See :meth:`do_test`."""
+        self.do_test('streamed')
+
+    def do_test(self, download_policy):
+        """Publish repository using different  ``download_policy``."""
+        repo = self.client.post(REPO_PATH, gen_repo())
+        self.addCleanup(self.client.delete, repo['_href'])
+
+        body = gen_file_remote(**{'policy': download_policy})
```
rochacbruno: why use `**` to explode this dict instead of just `gen_file_remote(policy=download_policy)` ?

kersommoura: I can update this. Thanks.

pulp_file/tests/functional/api/test_download_policies.py
```diff
@@ -86,3 +88,44 @@ def do_test(self, download_policy):
         self.assertNotEqual(latest_version_href, repo['_latest_version_href'])
         self.assertDictEqual(get_content_summary(repo), FILE_FIXTURE_SUMMARY)
         self.assertDictEqual(get_added_content_summary(repo), {})
+
+
+class PublishDownloadPolicyTestCase(unittest.TestCase):
```
dralley: Since this test class is nearly identical to SyncDownloadPolicyTestCase, do you think there would be value in merging them? I can see why keeping them separate is good too, it's just a thought.

kersommoura: @dralley, I have been thinking about merge the sync/publish tests. Since they are very similar code wise, but with different concerns. I will follow up with @pulp/qe  and update this.

rochacbruno: +1

then have `do_sync` and `do_publish`

kersommoura: @rochacbruno, changes in place.

PR #181, opened 2019-02-27
Sync in additive mode by default

goosemania:
re #4449
https://pulp.plan.io/issues/4449


kersommoura:
I was expecting this change to break a few tests. Interesting.

goosemania:
@kersommoura , me too, that's why I'm glad that https://pulp.plan.io/issues/4448 is filed :)

PR #196, opened 2019-03-27
Bumps version for beta 10

dkliban:
This beta is compatible with the RC 1 of pulpcore

[noissue]


daviddavis:
Oh something I just thought of—do we want to check in migrations?

daviddavis:
Also, after talking on IRC, I think we need to pin to `~=0.1.0rc1`.

daviddavis:
Trying to coax Travis by re-opening this PR.

PR #200, opened 2019-04-02
Add test to compare API bindings

kersommoura:
Use the generated OpenAPI schema in json format generated on
April/01/2019 as base to compare with generated on the fly API schemas.
Main goal is to detect changes in the API schema.

In case of intentional changes, the `stored_open_api.json` has to be
updated.

https://pulp.plan.io/issues/4123
closes #4123


pulp_file/tests/functional/api/test_api_schema.py
```diff
@@ -0,0 +1,76 @@
...
+
+    def test_verify_api_schema(self):
+        """Verify API schema change."""
+        cfg = config.get_config()
+        client = api.Client(cfg, api.json_handler)
+        generated_open_api = client.get(API_SCHEMA_PATH)
+
+        current_dir = os.path.dirname(os.path.abspath(__file__))
+        assets_dir = os.path.join(current_dir, 'assets')
+        temp_dir = os.path.realpath(tempfile.mkdtemp())
```
rochacbruno: I would replace it with

```python
with tempfile.TemporaryDirectory() as tmpdirname:
    # do the testing here
```

Then the entire tempdir will be deleted after the test finishes and no need to `addCleanup` for it.

kersommoura: I will update properly.

PR #204, opened 2019-04-10
Add test to verify download of content

kersommoura:
Perform a lazy sync, and change to immediate to force download.

Perform an immediate sync to download artifacts for content units that
are already created.

https://pulp.plan.io/issues/4467
closes:#4467

### DIFF
--- a/pulp_file/tests/functional/api/test_download_policies.py
+++ b/pulp_file/tests/functional/api/test_download_policies.py
@@ -1,10 +1,16 @@
 # coding=utf-8
 """Tests for Pulp`s download policies."""
+from random import choice
 import unittest
 
 from pulp_smash import api, config
-from pulp_smash.pulp3.constants import REPO_PATH
+from pulp_smash.pulp3.constants import (
+    ARTIFACTS_PATH,
+    LAZY_DOWNLOAD_POLICIES,
+    REPO_PATH,
+)
 from pulp_smash.pulp3.utils import (
+    delete_orphans,
     gen_repo,
     get_added_content_summary,
     get_content_summary,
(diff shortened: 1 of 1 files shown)


dralley:
If there's not a compelling reason to make it a per-plugin test, I would say the best place for it is in pulpcore "using_plugin" since it's testing core functionality.

bmbouter:
What is the future of "using_plugin" In talking with @rochacbruno it sounded like maybe it was going away in favor of plugin-tests that were also run in various other repos. For example the pulp-certguard functional tests are committed to pulp/certguard but run when pulp/pulpcore runs also.

kersommoura:
After further discussion with @bmbouter, lets keep this test here. If necessary add the same test for other plugins, if this is going to assure no regressions in the plugin itself.

@rochacbruno, ready to be reviewed.

pulp_file/tests/functional/api/test_download_policies.py
```diff
@@ -170,3 +176,47 @@ def do_test(self, policy):
         # feed.
         content = self.client.get(FILE_CONTENT_PATH)
         self.assertEqual(len(content), FILE_FIXTURE_COUNT, content)
+
+
+class ContentUnitstDownloadTestCase(unittest.TestCase):
```
rochacbruno: When I read `download` it gave me impression that we were downloading content from pulp, but this test is actually `syncing` content from remote.

As we are now downloading from Pulp Content app, what if we rename it to `ContentUnitsSyncTestCase` ?

kersommoura: @rochacbruno, naming is so hard.  I changed the name. Thanks.

dralley:
I don't really understand how there could be a regression in the plugin itself that would affect download, but I'm ok with this if you feel it's best.

PR #209, opened 2019-04-30
Problem: pulp_file_client not published as ruby gem

dkliban:
Solution: publish pulp_file_client ruby gem when building project

re: #4695
https://pulp.plan.io/issues/4695


.travis/credentials.enc
```diff
@@ -0,0 +1 @@
+�KH�5(�qM|V2HM��G}�\`۝�E�^��O��M�cW��3~[-11eq�r�?É�6$��`�
```
bmbouter: Should this file be more clearly named to be associated w/ ruby? I'm concerned about confusion when we add pypi support, etc, etc.

dkliban: The pypi credentials are already stored as a secure variable. So there is not going to be another credentials file for now.

PR #210, opened 2019-04-30
Problem: pulp not provisioned during gem publish step

dkliban:
Solution: configure travis environment for gem publish step

re: #4695
https://pulp.plan.io/issues/4695

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,26 +3,36 @@ sudo: required
 dist: xenial
 language: python
 python:
-    # python versions used in el7 SCL & supported fedora
-    - "3.6"
-    - "3.7"
-addons:
-    # postgres versions provided by el7 RHSCL (lowest supportable version)
-    postgresql: "9.6"
+  - '3.6'
+  - '3.7'
+env:
+  matrix:
+    - DB=postgres TEST=pulp
+    - DB=mariadb TEST=pulp
+    - DB=postgres TEST=bindings
+matrix:
+  exclude:
+    - python: '3.6'
+      env: DB=mariadb TEST=pulp
+    - python: '3.7'
+      env: DB=postgres TEST=bindings
+  fast_finish: true
 services:
-    - postgresql
-    - redis-server
-install: source .travis/install.sh
-before_script: source .travis/before_script.sh
-script: source .travis/script.sh
-after_failure:
-  - sh -c "cat ~/django_runserver.log"
-  - sh -c "cat ~/resource_manager.log"
-  - sh -c "cat ~/reserved_workers-1.log"
-stages:
-- name: test
-- name: deploy
-  if: tag IS present
+  - postgresql
+  - mariadb
+addons:
+  apt:
+    packages:
+      - httpie
+      - jq
+  # postgres versions provided by el7 RHSCL (lowest supportable version)
+  postgresql: '9.6'
+  mariadb: '10.3'
+before_install: .travis/before_install.sh
+install: .travis/install.sh
+before_script: .travis/before_script.sh
+script:
+  - .travis/script.sh
 jobs:
   include:
   - stage: deploy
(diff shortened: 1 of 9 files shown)


.travis.yml
```diff
@@ -35,6 +45,9 @@ jobs:
         secure: pQVElIk7hM0cyyGz8K+0R23OlcSBocYibi7yOPYUvQ/x8oTEuH/WFt18M5i3YK2tPg3Mypnm+uaUg1vxd+QlJRZTTTj+sSIiGflTkTMIWPpoQ9bm8/oo5MlOP630mK3/Lp117BLAqNJfUm6XW8dikhBFiWZ+IVge3bsuaIOTs+8TNmXT6F0aZZ66RQElcfdllfGSmYv39QzH95XL6d/AgSBh93he26mqlOpHf7gcsT/5UwggW8wDjyF53WYBBuYjg3A7umwHzoilLy98E3ortGF1RaMjlu0jSrlsT+pBdu1ZKwgxEfk3zqKaTYNr/0w9AudsRWMXY8F2oAOG1W5qUoabqiFiVc4AvrP2oZ6q3y1ikv1kLSjGpfmPdz524XY25xoXZOp2M5NIuVo36bZZ9tUhNtkGnE0pufzaRYDAAChWW5xvcOU1ngSDfFkogg4cUmWkVrARBhRwaUcbN2fU/lPDpdaPjg0UCLTCmO8VsGfdvX0NAGu8MUQjbOLvPqzAIR+pEDgKiRIqgW8UlByIAmjh7Z0W8NXLGcTNt2vP9qXnSJpjXP6YU7WlfFEQn9sFwIBQEiCiU2vUfV83iNoVDSaN+0zeeJCHmrY5g/FaqHvXw++vu0ZzK43ZLDy6TU58QP2qooZ6pie8/PGYWVr6U3avXtk/epC3hSHwo6tv1GA=
       on:
         tags: true
+    if: tag IS present
```
daviddavis: Isn't this handled already on the previous line?

.travis/before_install.sh
```diff
@@ -0,0 +1,64 @@
...
+if [ -n "$PULP_PLUGIN_PR_NUMBER" ]; then
+  pushd pulpcore-plugin
+  git fetch origin +refs/pull/$PULP_PLUGIN_PR_NUMBER/merge
+  git checkout FETCH_HEAD
+  popd
+fi
+
+
+if [ -n "$PULP_SMASH_PR_NUMBER" ]; then
+  pip uninstall -y pulp-smash
```
daviddavis: Why is this needed?

.travis.yml
dkliban: line 48 is needed. not exactly sure about line 47 though. but the docs say it's needed.

.travis/install.sh
```diff
@@ -1,44 +1,17 @@
...
-
-git clone https://github.com/pulp/pulpcore-plugin.git
-
-if [ -n "$PULP_PLUGIN_PR_NUMBER" ]; then
-  pushd pulpcore-plugin
-  git fetch origin +refs/pull/$PULP_PLUGIN_PR_NUMBER/merge
-  git checkout FETCH_HEAD
-  popd
+if [ "$TEST" = 'docs' ]; then
+  pip3 install -r doc_requirements.txt
```
daviddavis: Why are you using pip3 here? Asking because everywhere else uses pip.

.travis/before_install.sh
dkliban: copy/pasta from pulpcore. this is not needed. i'll remove.

.travis/install.sh
dkliban: copy/pasta from pulpcore. i can update.

.travis/script.sh
```diff
@@ -1,30 +1,61 @@
...
 
 # Lint code.
-flake8 --config flake8.cfg || exit 1
-
-# Run migrations.
-export DJANGO_SETTINGS_MODULE=pulpcore.app.settings
-export PULP_CONTENT_HOST=localhost:24816
-django-admin makemigrations file
-django-admin migrate --noinput
+flake8 --config flake8.cfg
```
daviddavis: I'm a bit hesitant about dropping the `|| exit 1` since a faster fail time would help alleviate Travis traffic. Thoughts?

dkliban: let's fail fast

daviddavis:
Did you intend to add testing against mariadb? Asking because I don't see it mentioned in the commit message at all.

dkliban:
> Did you intend to add testing against mariadb? Asking because I don't see it mentioned in the commit message at all.

Yes I did mean to. I'll add it to the commit message.

PR #219, opened 2019-05-09
Use PublicationDistribution from pulpcore.plugin

bmbouter:
The plugin API now uses PublicationDistribution which causes pulp_file
users to only receive the fields they can actually use.

Required PR: https://github.com/pulp/pulpcore/pull/124
Required PR: https://github.com/pulp/pulpcore-plugin/pull/97

https://pulp.plan.io/issues/4785
closes #4785


pulp_file/app/serializers.py
```diff
@@ -81,22 +81,31 @@ class FilePublicationSerializer(PublicationSerializer):
     Serializer for File Publications.
     """
 
+    _file_distributions = DetailRelatedField(
```
gmbnomis: What do you think about just using the default relation name (i.e. no need for `default_related_name`) and make the field available at `_distributions` in the serializer:

```python
    _distributions = RelatedField(
        help_text=_(
            "This publication is currently being served as defined by these distributions."
        ),
        source="filedistribution_set",
        many=True,
        read_only=True,
        view_name="publications-file/file-detail",
    )
```

As an alternative we could use a `DetailRelatedField`. It is simpler to use, although not really necessary in this case, since we don't need to support subclasses of FileDistribution.

asmacdo: Since this is a plugin defined field, should we drop the underscore?

gmbnomis: Good question. Does a convention that should be implemented the same in all plugins make the field a "core field"? (we had a slightly similar discussion on `_artifact` once, IIRC). I am fine either way and slightly preferring to drop the underscore.

bmbouter: From the feedback I think the best thing to do is to switch this to be `distributions`. That both makes it a core variable and also brings more consistency with the hope that more plugins will do the same. Does this sound right?

gmbnomis: yes, PR looks good to me now

PR #226, opened 2019-05-17
Expand and clarify rendered docstrings

asmacdo:


flake8.cfg
```diff
@@ -1,6 +1,6 @@
 [flake8]
 exclude = ./docs/*,*/migrations/*
-ignore = Q000,D100,D104,D106,D200,D401,D402
+ignore = Q000,Q003,D100,D104,D106,D200,D205,D400,D401,D402
```
asmacdo: Our docstring linting is focused on developers, and since our docstrings do double duty (they become user-facing documentation), the format needs to be less restrictive.

pulp_file/app/viewsets.py
```diff
@@ -53,7 +55,11 @@ class FileContentViewSet(ContentViewSet):
 
 class FileRemoteViewSet(RemoteViewSet):
     """
-    ViewSet for File Remotes.
+    <!-- User-facing documentation, rendered as html-->
+    FileRemote represents an external source of <a
+    href="../restapi.html#operation/content_file_files_list">File Content</a>.  The target url of a
```
dkliban: the href is wrong when I am looking at the live docs. It ends up as /pulp/api/v3/restapi.html#operation/content_file_files_list

dkliban: I think you can just change the href to be "#operation/content_file_files_list" and it will always work

PR #233, opened 2019-05-22
Adding checksums to file attributes

daviddavis:
fixes #4827
https://pulp.plan.io/issues/4827

### DIFF
--- a/pulp_file/app/viewsets.py
+++ b/pulp_file/app/viewsets.py
@@ -46,7 +46,7 @@ class FileContentViewSet(ContentViewSet):
     """
 
     endpoint_name = 'files'
-    queryset = FileContent.objects.all()
+    queryset = FileContent.objects.prefetch_related("_artifacts")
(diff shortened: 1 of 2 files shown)


pulp_file/app/viewsets.py
```diff
@@ -46,7 +46,7 @@ class FileContentViewSet(ContentViewSet):
     """
 
     endpoint_name = 'files'
-    queryset = FileContent.objects.all()
+    queryset = FileContent.objects.prefetch_related("_artifacts").all()
```
dralley: Should it be the other way around?  

```FileContent.objects.all().prefetch_related("_artifacts")````

daviddavis: I tested it and couldn't detect a difference. I shortened it to `FileContent.objects.prefetch_related("_artifacts")` to avoid confusion and to make it simpler. Let me know if you disagree with that.

PR #249, opened 2019-06-27
Add black to pulp_file

daviddavis:
fixes #5021

### DIFF
--- a/flake8.cfg
+++ b/flake8.cfg
@@ -1,10 +1,11 @@
 [flake8]
 exclude = ./docs/*,*/migrations/*
-ignore = Q000,Q003,D100,D104,D106,D200,D205,D400,D401,D402
+ignore = W503,Q000,Q003,D100,D104,D106,D200,D205,D400,D401,D402,D202
 max-line-length = 100
 
 # Flake8-quotes extension codes
 # -----------------------------
+# W503: This enforces operators before line breaks which is not pep8 or black compatible.
 # Q000: double or single quotes only, default is double (don't want to enforce this)
 
 # Flake8-docstring extension codes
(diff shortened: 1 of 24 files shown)


flake8.cfg
```diff
@@ -1,10 +1,11 @@
 [flake8]
 exclude = ./docs/*,*/migrations/*
-ignore = Q000,Q003,D100,D104,D106,D200,D205,D400,D401,D402
+ignore = W503,Q000,Q003,D100,D104,D106,D200,D205,D400,D401,D402,D202
 max-line-length = 100
 
 # Flake8-quotes extension codes
 # -----------------------------
+# W503: Places operators before line breaks. Not pep8 or black compatible.
```
bmbouter: I don't understand this comment. The non-ignoring of W503 is not pep8 or black compatible?

daviddavis: W503 is not pep-8 or black compatible. [Pep-8 allows for both and even seems to suggest the linebreak should go before the operator](https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator). See https://github.com/python/black/issues/52 for more info.

Would "Places operators before line breaks and is not pep8 or black compatible." be more clear?

bmbouter: I think the confusion is that since they are two sentences it's not clear if placing operators before line breaks is compatible or incompatible. I read more of the links so I know the difference, but something like this would be the clearer to me:

W503 ignoring W503 places operators before line breaks. Not ignoring is incompatible with pep8 and black."

Is what I'm saying correct?

daviddavis: Your sentence is a bit ambiguous so I am not sure if you're correct or not. W503 enforces the placement of operators before line breaks. This is not pep-8 or black compatible. 

TBH, the way this file is written is difficult to understand to me but this sentence does seem to fit its pattern in that it the first part explains what the rule enforces and the second part describes why we're disabling it.

daviddavis: How about "W503: This enforces operators before line breaks which is not pep8 or black compatible."

bmbouter: This last proposal reads super clear to me.

PR #272, opened 2019-09-09
adding code on progress bar

fabricio-aguiar:
related to: https://github.com/pulp/pulp_file/pull/272
ref #5184
https://pulp.plan.io/issues/5184

### DIFF
--- a/CHANGES/5184.feature
+++ b/CHANGES/5184.feature
@@ -1,0 +1,1 @@
+Setting `code` on `ProgressBar`.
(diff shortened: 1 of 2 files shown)


CHANGES/5184.misc
```diff
@@ -0,0 +1 @@
+Adding `code` on `ProgressBar`.
```
daviddavis: Nitpick should probably be "Setting code..." rather than "Adding code...".

Also, I'd probably consider this a feature?

PR #285, opened 2019-10-02
0.1.0b4.dev

daviddavis:
[noissue]

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,14 +2,14 @@
 
 from setuptools import setup, find_packages
 
-requirements = ["pulpcore-plugin==0.1.0rc6"]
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -2,14 +2,14 @@
 
 from setuptools import setup, find_packages
 
-requirements = ["pulpcore-plugin==0.1.0rc6"]
+requirements = ["pulpcore-plugin~=0.1.0rc6"]
```
goosemania: maybe ~=0.1rc6 is better for consistency across plugins?
I'm not sure if ~=0.1.0rc6 covers 0.1.0rc7 but I haven't checked.

PR #286, opened 2019-10-03
Updating travis

fabricio-aguiar:
[noissue]


fabricio-aguiar:
@bmbouter it is green now!

PR #290, opened 2019-10-07
Prevent duplicate `relative_path` files

bmbouter:
When a RepositoryVersion has another `file.file` added with the same
`relative_path` the original one should be removed. This is shown
correctly on the RepositoryVersion `added`, `removed` fields.

This adds a test which uploads two distinct Artifacts, makes them into
File content, and then associates one, and then another with a Repo. The
second RepositoryVersion should have only 1 file.file content unit in
it.

Required PR: https://github.com/pulp/pulpcore/pull/331

https://pulp.plan.io/issues/4028
closes #4028

### DIFF
--- a/pulp_file/tests/functional/api/test_crud_content_unit.py
+++ b/pulp_file/tests/functional/api/test_crud_content_unit.py
@@ -6,10 +6,10 @@ from requests.exceptions import HTTPError
 
 from pulp_smash import api, config, utils
 from pulp_smash.exceptions import TaskReportError
-from pulp_smash.pulp3.constants import ARTIFACTS_PATH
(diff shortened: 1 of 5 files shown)


pulp_file/tests/functional/api/test_crud_content_unit.py
```diff
@@ -233,3 +233,57 @@ def test_non_error(self):
 
         # create second content unit.
         self.client.post(FILE_CONTENT_PATH, gen_file_content_attrs(artifact))
+
+
+class DuplicateRelativePathsInRepo(unittest.TestCase):
```
daviddavis: :+1:

bmbouter:
@daviddavis I removed that other logic, lmk what you think.

PR #294, opened 2019-10-15
Verson bump to 0.1.0b5.dev

bmbouter:
* Now depends on pulpcore-plugin>=0.1.0rc7

[noissue]

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,14 +2,14 @@
 
 from setuptools import setup, find_packages
 
-requirements = ["pulpcore-plugin==0.1rc7"]
+requirements = ["pulpcore-plugin>=0.1rc7"]
 
 with open("README.rst") as f:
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -2,14 +2,14 @@
 
 from setuptools import setup, find_packages
 
-requirements = ["pulpcore-plugin==0.1rc7"]
+requirements = ["pulpcore-plugin>=0.1rc7"]
```
bmbouter: Should this be set like this? My concern was this it should work with the existing rc until we have to pin it to the next one. Once it's pinned we can't test this on top of the GA pypi bits was my concern.

daviddavis: We've been using `~=`

dkliban: This looks fine to me.

PR #299, opened 2019-10-28
Typed Repositories

dralley:


pulp_file/app/models.py
```diff
@@ -31,7 +31,19 @@ class Meta:
...
 
 
-class FileRemote(Remote):
+class FileRepository(core_models.Repository):
+    """
+    The "file" repository type.
+    """
+
+    TYPE = "file"
+    CONTENT_TYPES = [FileContent]
```
goosemania: @dralley , where/how are those used?

dralley: Currently, nowhere.  The information will probably be useful in the stages API but it isn't actually used for anything yet.  We can skip it if you want to.

gmbnomis:
While working on pulp_cookbook I noticed that the link to the repository API (in this PR: https://github.com/pulp/pulp_file/pull/299/files#diff-4ea13541b01c775a55f25a8c4fab0aebR130) needs to be fixed, as it moved from core to here.

dralley:
@gmbnomis I tried to fix it and got a docs build error, so given the time constraints, I'll just take note of that and get the docs fixed after the dev freeze.

gmbnomis:
> @gmbnomis I tried to fix it and got a docs build error, so given the time constraints, I'll just take note of that and get the docs fixed after the dev freeze.

no prob. Fyi, here is my change for pulp cookbook: https://github.com/gmbnomis/pulp_cookbook/pull/64/files#diff-8e32a264a0c2f46c5aa1ff6240de6feeR135 (I generally changed the links to operations into links to endpoints)

PR #307, opened 2019-11-07
Sync, Upload, Modify preserve uniqueness

bmbouter:
Sync, Upload, Modify preserve uniqueness

Sync, Upload, and Modify now have added content with the same
`relative_path` as existing content will remove the existing content.

Required PR: https://github.com/pulp/pulpcore/pull/369

https://pulp.plan.io/issues/3541
re #3541

### DIFF
--- a/pulp_file/app/viewsets.py
+++ b/pulp_file/app/viewsets.py
@@ -1,8 +1,7 @@
-from gettext import gettext as _  # noqa
-
 from drf_yasg.utils import swagger_auto_schema
 from rest_framework.decorators import action
 
+from pulpcore.plugin.actions import ModifyRepositoryActionMixin
 from pulpcore.plugin.serializers import (
     AsyncOperationResponseSerializer,
     PublicationExportSerializer,
@@ -63,7 +62,7 @@ class FileContentViewSet(SingleArtifactContentUploadViewSet):
     filterset_class = FileContentFilter
 
 
-class FileRepositoryViewSet(RepositoryViewSet):
+class FileRepositoryViewSet(RepositoryViewSet, ModifyRepositoryActionMixin):
     """
     <!-- User-facing documentation, rendered as html-->
     FileRepository represents a single file repository, to which content can be synced, added,
--- a/CHANGES/3541.feature
+++ b/CHANGES/3541.feature
@@ -1,0 +1,2 @@
+Sync, Upload, and Modify now have added content with the same `relative_path` as existing content
+will remove the existing content.
--- a/pulp_file/app/models.py
+++ b/pulp_file/app/models.py
@@ -10,6 +10,7 @@ from pulpcore.plugin.models import (
     Remote,
     Repository,
 )
+from pulpcore.plugin.repo_version_utils import remove_duplicates
 
 
 log = getLogger(__name__)
@@ -28,7 +29,7 @@ class FileContent(Content):
     """
 
     TYPE = "file"
-    repo_key = ("relative_path",)
+    repo_key_fields = ("relative_path",)
 
     relative_path = models.TextField(null=False)
     digest = models.CharField(max_length=64, null=False)
@@ -47,6 +48,17 @@ class FileRepository(Repository):
 
     class Meta:
         default_related_name = "%(app_label)s_%(model_name)s"
+
+    def finalize_new_version(self, new_version):
+        """
+        Ensure no added content contains the same `relative_path` as other content.
+
+        Args:
+            new_version (pulpcore.app.models.RepositoryVersion): The incomplete RepositoryVersion to
+                finalize.
+
+        """
+        remove_duplicates(new_version)
 
 
 class FileRemote(Remote):


pulp_file/app/tasks/synchronizing.py
```diff
@@ -41,8 +42,10 @@ def synchronize(remote_pk, repository_pk, mirror):
         raise ValueError(_("A remote must have a url specified to synchronize."))
 
     first_stage = FileFirstStage(remote)
-    dv = DeclarativeVersion(first_stage, repository, mirror=mirror)
-    dv.create()
+    with RepositoryVersion.create(repository) as new_version:
+        dv = DeclarativeVersion(first_stage, new_version, mirror=mirror)
+        dv.create()
+        RemoveDuplicates(new_version).run()
```
dralley: I really like this.  I bet it makes custom pipelines a lot easier too.

gmbnomis: I would have expected a pattern like `RemoveDuplicates().run(new_version)` . If we were to pass around validators/modifiers in the future, this would allow to prepare validator instances (with config options) and use them in generic create/add_remove tasks. wdyt?

bmbouter: +1 to this I'm going to make this change. It seems more flexible for the caller to have the reference to `new_version` for the generic create/add_remove tasks as you mention.

pulp_file/app/serializers.py
```diff
@@ -49,6 +50,22 @@ def deferred_validate(self, data):
...
+        repository = validated_data.pop("repository", None)
+        content = super().create(validated_data)
+
+        if repository:
+            content_to_add = self.Meta.model.objects.filter(pk=content.pk)
+
+            # create new repo version with uploaded package
+            with repository.new_version() as new_version:
+                new_version.add_content(content_to_add)
+                RemoveDuplicates().run(new_version)
```
bmbouter: With the `new_version` passed in as part of `run(new_version)`, this could work differently. We could introduce an optional param on `SingleArtifactContentUploadSerializer.create` in core called plugin_modify_validate which the plugin could pass in.

That would make this function instead:

```
def create(self, validated_data):
    content = super().create(validated_data, plugin_modify_validate=RemoveDuplicates())
```

dkliban: @bmbouter This would be an improvement over what we currently have. Let's do it.

bmbouter: @gmbnomis this was made possible by your suggestion to have the called of `run()` provide `new_version`. What do you think of me making ^ change?

goosemania: @bmbouter, how will you call the plugin_modify_validate in core? `plugin_modify_validate.run(new_version)`?
I'm just trying to understand the interface if I have multiple modifiers/validators in plugin. I guess I can combine all of them in one (if they can run after content was added).

bmbouter: Yes that's how I imagined it would be called. It's only prepared to call a single item so the plugin writer would wrap multiple ones together. Another option would be to have a list or optionall have `plugin_modify_validate` accept either a single item or a list of items to call in order.

goosemania: I think it's fine to call one combined one and if plugin needs to do something in between, it has flexibility to do so.

+1 to your proposal

gmbnomis: @bmbouter although it is "hooky", I like the pattern. The hook has a well defined boundary which is rather independent of how `create` is actually implemented. (the previous solution could break easily  with changes to `create` in pulpcore).

However, `create()` now takes a different number of arguments depending on which class it is defined in, which isn't very nice. Should we add an instance member that stores the validator/modifier instance (with a class member that defaults to None)? Then, one would simply set it e.g. in the subclass (class member), in an overloaded __init__ or from the outside before calling create (i.e. the pattern that you see all the time in Django/DRF).

bmbouter: @gmbnomis that is a great idea, let me put that together. Do you have a suggestion for the class member name?

pulp_file/app/tasks/synchronizing.py
```diff
@@ -41,7 +42,7 @@ def synchronize(remote_pk, repository_pk, mirror):
         raise ValueError(_("A remote must have a url specified to synchronize."))
 
     first_stage = FileFirstStage(remote)
-    dv = DeclarativeVersion(first_stage, repository, mirror=mirror)
+    dv = DeclarativeVersion(first_stage, repository, mirror=mirror, vadifier=RemoveDuplicates().run)
```
mdellweg: Did you mean `validifier` or `validator`?

pulp_file/app/viewsets.py
```diff
@@ -63,7 +65,14 @@ class FileContentViewSet(SingleArtifactContentUploadViewSet):
     filterset_class = FileContentFilter
 
 
-class FileRepositoryViewSet(RepositoryViewSet):
+class FileVadifiyModifyActionMixin(ModifyActionMixin):
+
+    def _dispatch_task(self, *args, **kwargs):
```
mdellweg: This `_dispatch_task` is not called on sync, right?
What tasks is it called on? I believe there is not much use on delete.
Also should we really pass inflated objects to the task queue?

pulp_file/app/tasks/synchronizing.py
bmbouter: I meant both because it could perform either. The feedback is that folks want a different name so we're going to rename it once we determine what a better name is.

pulp_file/app/viewsets.py
bmbouter: > This `_dispatch_task` is not called on sync, right?
Correct, it's called as part of the `modify` endpoint. With typed repos merged, [this endpoint](https://github.com/pulp/pulpcore/blob/2161bf71ddce76394933fe0d52b09194a2153501/pulpcore/app/viewsets/repository.py#L98-L136) was provided on all Repository viewsets, but now it's an opt-in endpoint with plugin writers adding the mixin to their Detail RepositoryViewset, e.g. L#68.

> What tasks is it called on? I believe there is not much use on delete.
This is in the viewset, but it's going to call the [`add_and_remove` task](https://github.com/pulp/pulpcore/blob/fa39253c93c324142c7459a46a90753487d075db/pulpcore/app/tasks/repository.py#L69)

> Also should we really pass inflated objects to the task queue?
This is also a concern of mine, but I'm not totally sure why other than pickling non-native Python types makes me nervous. RQ pickles its arguments so I'm not concerned with it working. I am a little concerned though.

PR #312, opened 2019-11-14
Problem: We cannot see pulpcore downgrades during plugin installs

mikedep333:
because the containers build.yaml output is suppressed unless
the ansible task errors.

Solution: Regenerate with plugin-template

[noissue]


mikedep333:
@daviddavis or someone else: Please merge. I cannot merge.

Also review (and merge while you're at it):
https://github.com/pulp/plugin_template/pull/141/files

PR #313, opened 2019-11-14
Create version 0.1.0rc1

bmbouter:


setup.py
```diff
@@ -2,14 +2,14 @@
 
 from setuptools import setup, find_packages
 
-requirements = ["pulpcore>=3.0rc7"]
+requirements = ["pulpcore>=3.0rc8,pulpcore<3.1"]
```
evgeni:
```suggestion
requirements = ["pulpcore>=3.0rc8,<3.1"]
```

evgeni: (that's why the CI is red)

bmbouter: I'm applying this change. Thank you for pointing this out.

PR #317, opened 2019-11-19
use HTTPS for url in setup.py

evgeni:
also drop www, as it redirects anyways


bmbouter:
@evgeni same for this one w/ the `[noissue]` needed.

PR #324, opened 2019-11-25
use HTTPS for url in setup.py

bmbouter:
also drop www, as it redirects anyways

[noissue]

(cherry picked from commit 3b8367ce407939e2c619529469f2e704148d5e4c)


bmbouter:
I'm going to make a new cherry pick PR now

PR #329, opened 2019-12-05
Add CONTENT_TYPES to repo definition

dralley:
re: #5701
https://pulp.plan.io/issues/5701

### DIFF
--- a/CHANGES/5701.misc
+++ b/CHANGES/5701.misc
@@ -1,0 +1,1 @@
+Added CONTENT_TYPES to the FileRepository definition.
--- a/pulp_file/app/models.py
+++ b/pulp_file/app/models.py
@@ -46,6 +46,7 @@ class FileRepository(Repository):
     """
 
     TYPE = "file"
+    CONTENT_TYPES = [FileContent]
 
     class Meta:
         default_related_name = "%(app_label)s_%(model_name)s"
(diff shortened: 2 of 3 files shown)


pulp_file/tests/unit/test_core/test_repo_version.py
```diff
@@ -0,0 +1,52 @@
+from unittest.mock import patch
+
+from django.core.files.uploadedfile import SimpleUploadedFile
```
dralley: It's tempting to just get rid of this entire test, since it's thoroughly covered by functional tests, and we're technically duplicating effort.  But it's already here and it already works and it takes a fraction of a second to run so we may as well keep it I guess?  What do you think.

pulp_file/app/models.py
```diff
@@ -46,6 +46,7 @@ class FileRepository(Repository):
     """
 
     TYPE = "file"
+    CONTENT_TYPES = [FileContent]
```
bmbouter: Will merging the pulpcore PR break if this is not merged also?

CHANGES/5701.misc
```diff
@@ -0,0 +1 @@
+Added CONTENT_TYPES to the FileRepository definition.
```
bmbouter: Can a newline character be added at the end here?

pulp_file/tests/unit/test_core/test_repo_version.py
bmbouter: It's ok to leave since it's written and it works.

pulp_file/app/models.py
dralley: Yup.  But this one can be merged without it despite the "required PR" tag.

PR #330, opened 2019-12-09
Increase sync speed by saving less progress reports

dralley:
I don't have numbers on this for the File plugin but for RPM it sped things up by at least 30%.  One database save per file (```.increment()```) is a huge cost.

[noissue]

### DIFF
--- a/pulp_file/app/tasks/synchronizing.py
+++ b/pulp_file/app/tasks/synchronizing.py
@@ -19,6 +19,17 @@ from pulp_file.manifest import Manifest
 log = logging.getLogger(__name__)
 
 
+def grouper(iterable, n):
+    """
+    Generic batcher over iterators.
+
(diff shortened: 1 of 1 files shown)


dralley:
Black and flake8/pep8speaks are conflicting regarding what style is correct, if I fix it for one the other breaks.  We should probably pick one and disable the others.

Or, ignore ```E203``` in our linting configs.

daviddavis:
Black calls out this problem: 

> This behaviour may raise E203 whitespace before ':' warnings in style guide enforcement tools like Flake8. Since E203 is not PEP 8 compliant, you should tell Flake8 to ignore these warnings.

And there's [a flake8 issue where they're asking for a PR](https://github.com/PyCQA/pycodestyle/issues/373).

Here's a plugin_template PR: https://github.com/pulp/plugin_template/pull/152

PR #335, opened 2019-12-16
Assuring downloading ISO does not go to stdout

fabricio-aguiar:
https://pulp.plan.io/issues/5775
closes #5775

Example:
https://travis-ci.com/pulp/pulp-operator/jobs/267613747#L2387-L14676

### DIFF
--- a/CHANGES/5775.misc
+++ b/CHANGES/5775.misc
@@ -1,0 +1,1 @@
+Assuring downloading ISO does not go to stdout during tests.
--- a/docs/_scripts/download_after_sync.sh
+++ b/docs/_scripts/download_after_sync.sh
@@ -11,4 +11,4 @@ fi
 
 # Next we download a file from the distribution
 # This will default to http://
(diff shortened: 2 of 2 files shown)


CHANGES/5775.misc
```diff
@@ -0,0 +1 @@
+Assuring downloading ISO does not go to stdout.
```
mikedep333: Should we clarify this is for the test scripts?

bmbouter:
I can see how this is good for the docs, but I don't see where download_after sync is called at test time https://github.com/pulp/pulp_file/search?q=download_after_sync&unscoped_q=download_after_sync

fabricio-aguiar:
> I can see how this is good for the docs, but I don't see where download_after sync is called at test time https://github.com/pulp/pulp_file/search?q=download_after_sync&unscoped_q=download_after_sync

they are called here:
https://github.com/pulp/pulp-operator/blob/master/.travis/pulp_file-tests.sh

PR #341, opened 2020-01-10
Revert "Assuring downloading ISO does not go to stdout"

daviddavis:
This reverts commit 57a81241e778aaa874b1b3db3748e6c19edacdd5.

The change is causing the docs job to fail in Travis:

```
http: error: Cannot use --output, -o with redirected output.
```

refs #5775

Example of the failure: https://travis-ci.org/pulp/pulp_file/jobs/635423885

### DIFF
--- a/CHANGES/5775.misc
+++ b/CHANGES/5775.misc
@@ -1,1 +1,0 @@
-Assuring downloading ISO does not go to stdout during tests.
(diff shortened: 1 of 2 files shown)


daviddavis:
Actually I've filed https://pulp.plan.io/issues/5967 instead.

PR #342, opened 2020-01-15
Revert "Downloading ISO does not go to stdout"

fabricio-aguiar:
This reverts commit 57a81241e778aaa874b1b3db3748e6c19edacdd5.

closes #5967
https://pulp.plan.io/issues/5967

### DIFF
--- a/CHANGES/5775.misc
+++ b/CHANGES/5775.misc
@@ -1,1 +1,0 @@
-Assuring downloading ISO does not go to stdout during tests.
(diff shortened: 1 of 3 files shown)


CHANGES/5967.bugfix
```diff
@@ -0,0 +1 @@
+Revert "Downloading ISO does not go to stdout"
```
daviddavis: This should be a misc change I think.

PR #347, opened 2020-01-29
Functional tests using bindings

fabricio-aguiar:
[noissue]

### DIFF
--- a/pulp_file/tests/functional/api/test_crud_content_unit.py
+++ b/pulp_file/tests/functional/api/test_crud_content_unit.py
@@ -2,26 +2,25 @@
 """Tests that perform actions over content unit."""
 import unittest
 
-from requests.exceptions import HTTPError
-
-from pulp_smash import api, config, utils
-from pulp_smash.exceptions import TaskReportError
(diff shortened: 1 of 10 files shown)


pulp_file/tests/functional/api/test_crud_content_unit.py
```diff
@@ -122,52 +121,50 @@ class ContentUnitUploadTestCase(unittest.TestCase):
     @classmethod
     def setUpClass(cls):
         """Create class-wide variable."""
-        cls.cfg = config.get_config()
-        delete_orphans(cls.cfg)
+        delete_orphans()
         cls.content_unit = {}
-        cls.client = api.Client(cls.cfg, api.smart_handler)
-        cls.files = {"file": utils.http_get(FILE_URL)}
+        cls.api = ContentFilesApi(gen_file_client())
```
dralley: So this is fine as is, but what do you think about using a more descriptive name for this variable?  Like ```file_content_api``` instead of just ```api```.  That change would probably need to be made in the template.

PR #348, opened 2020-01-30
Run the pulp-fixtures container for the performance tests too

lubosmj:
[noissue]

### DIFF
--- a/.travis/before_script.sh
+++ b/.travis/before_script.sh
@@ -32,16 +32,15 @@ else
 fi
 
 
-# set up pulp-fixtures docker container
-if [[ "$TEST" == 'pulp' ]]; then
-  docker run -d -p 0.0.0.0:8000:80 quay.io/pulp/pulp-fixtures:latest
-fi
 cat ~/.config/pulp_smash/settings.json | \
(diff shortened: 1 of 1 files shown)


pulp_file/tests/performance/pulpperf/interact.py
```diff
@@ -20,16 +20,8 @@ def get(url, params={}):
...
-    while True:
-        data = get(url, params)
-        out += data["results"]
-        page += 1
-        params["page"] = page
-        if data["next"] is None:
-            break
-    return out
+    data = get(url, params)
+    return data["results"]
```
dralley: Why is this? Results are paginated so this will only get the first page of them (the first 100, or whatever)

lubosmj: Yay, I tested it against a different repository and there were not so many results returned. And because of this, it always failed with this error:

`400 Client Error: Bad Request for url: http://localhost:24817/pulp/api/v3/content/file/files/?repository_version=%2Fpulp%2Fapi%2Fv3%2Frepositories%2Ffile%2Ffile%2F67a56421-5834-4eff-98c6-cff24c2c658c%2Fversions%2F1%2F&page=1
`

So I thought that this part of the code is redundant. I restored the changes.

PR #350, opened 2020-01-31
Merge 0.1.1 to 0.1 branch

bmbouter:


bmbouter:
There are a variety of issues preventing this from passing on Travis. First it references commits in Dec that are in CLOSED CURRENT RELEASE because they are actually pulpcore issues. Travis won't pass because of that so I'm going to admin merge. That is safe because when I cut the 0.1.1 release it will be fully re-tested properly then.

PR #356, opened 2020-02-12
Including test scenario

fabricio-aguiar:
Depends on: https://github.com/pulp/pulpcore/pull/537
https://pulp.plan.io/issues/5567
closes #5567

### DIFF
--- a/CHANGES/5567.misc
+++ b/CHANGES/5567.misc
@@ -1,0 +1,1 @@
+Include test scenario: Content with duplicate repo_key_fields raises an error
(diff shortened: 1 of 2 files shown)


daviddavis:
You're hitting a different error. In pulpcore, we [validate that file locations in repo versions don't overlap or duplicate](https://github.com/pulp/pulpcore/blob/master/pulpcore/plugin/repo_version_utils.py#L64). This is independent of the repo_key_fields validation but unfortunately it means that you can't test this issue with file content since file content uses relative_path also as its repo key. 

I think you could use packages but you'd have to create two with the same [nevra](https://github.com/pulp/pulp_rpm/blob/master/pulp_rpm/app/models/package.py#L187) in pulp. I'm not sure if that's easy or if there's a fixture. Maybe module default or a comps content might be easier? Any suggestions @goosemania ?

daviddavis:
This test is good to have though as it tests our relative_path duplicate validation.

fabricio-aguiar:
> This test is good to have though as it tests our relative_path duplicate validation.

I changed the description

PR #357, opened 2020-02-14
Rename the filter for the field 'digest'

lubosmj:
closes #5965
https://pulp.plan.io/issues/5965


daviddavis:
@lubosmj I talked to Katello and they still need to filter by sha256. There's a sha256 field in the api. Could you add a `sha256` filter?

pulp_file/app/viewsets.py
```diff
@@ -43,10 +44,11 @@ class FileContentFilter(ContentFilter):
     """
     FilterSet for FileContent.
     """
+    sha256 = CharFilter(field_name="digest")
```
daviddavis: Nitpick but I would probably use `_artifacts__sha256` to be consistent with what other plugins will probably use.

daviddavis:
Went ahead and updated the code to get it ready for the 0.2.0. Thanks for the work!

https://github.com/pulp/pulp_file/pull/359

PR #360, opened 2020-02-26
0.2.0 version bumps and changelog

daviddavis:


setup.py
```diff
@@ -9,7 +9,7 @@
```
bmbouter: For this one the release commit is right because we want do declare maximum compat at 3.2, but for the last commit I think it should declare < 3.4 because otherwise when testing against a change that requires a master checkout of pulpcore it will be incompatible.

PR #362, opened 2020-02-27
Running tests on localstack

fabricio-aguiar:
https://pulp.plan.io/issues/6155
ref #6155

### DIFF
--- a/pulp_file/tests/functional/utils/__init__.py
+++ b/pulp_file/tests/functional/utils/__init__.py
@@ -6,6 +6,8 @@ from unittest import SkipTest
 from time import sleep
 from tempfile import NamedTemporaryFile
 
+from django.conf import settings
+
 from pulp_smash import api, selectors, utils
 from pulp_smash.pulp3.utils import (
     gen_remote,
@@ -25,6 +27,8 @@ from pulp_file.tests.functional.constants import (
     FILE_REPO_PATH,
     FILE_URL,
 )
+
+from pulp_file.tests.functional.utils.s3 import create_bucket
 
 from pulpcore.client.pulpcore import (
     ApiClient as CoreApiClient,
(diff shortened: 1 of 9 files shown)


fabricio-aguiar:
@bmbouter @daviddavis @dkliban could you please give me some feedback? Once I get an approval, I'll move these changes to plugin_template

dkliban:
The tests should run on PRs and Cron jobs. We should run jut the pulp_file tests.

pulp_file/tests/functional/utils/__init__.py
```diff
@@ -173,8 +177,10 @@ def monitor_task(task_href):
     completed = ["completed", "failed", "canceled"]
     task = tasks.read(task_href)
     while task.state not in completed:
-        sleep(2)
+        sleep(1)
         task = tasks.read(task_href)
+        if "S3Boto3Storage" in settings.DEFAULT_FILE_STORAGE:
+            create_bucket()
```
daviddavis: Why are we creating buckets here?

fabricio-aguiar: For minio, I have some problems with the connections pool:
urllib3.exceptions.ProtocolError: ('Connection aborted.', RemoteDisconnected('Remote end closed connection without response'))
travis: https://travis-ci.org/pulp/pulp_file/jobs/654626953?utm_medium=notification&utm_source=github_status
code: https://github.com/pulp/pulp_file/pull/358/files

For localstack, I can see the bucket has some lifetime, and it dies before the tests finish, so I added a conftest, and this method here, for keeping the bucket alive.

fabricio-aguiar: In theory:
- localstack bucket should not die, as I set `DATA_DIR`
- minio problem is fixed: https://github.com/minio/minio/pull/6415
But for both, I could not make it work locally and here on travis, the only way I could make localstack work was the trick of keeping creating bucket

daviddavis: There's no way to configure the lifetime of buckets in localstack? That also seems suboptimal for development as well if I have to create a bucket every so often.

fabricio-aguiar:
closing in favor of minio: https://github.com/pulp/pulp_file/pull/358

PR #372, opened 2020-04-07
Fixed non unique content units on content list

fabricio-aguiar:
https://pulp.plan.io/issues/6347
closes #6347


fabricio-aguiar:
Moving to pulpcore

PR #381, opened 2020-04-23
Documented pulp_file bindings client

fabricio-aguiar:
https://pulp.plan.io/issues/6518
closes #6518


daviddavis:
Thanks for the PR. How will these docs get updated whenever there is a change to the API? Will it be up to developers to remember to regenerate these docs when they make a change?

docs/pulp_file-client.rst
```diff
@@ -0,0 +1,306 @@
...
+   * - *ContentFilesApi*
+     - `\ **list** <pulp_file-client/ContentFilesApi.md#list>`_
+     - **GET** /pulp/api/v3/content/file/files/
+     - List file contents
+   * - *ContentFilesApi*
+     - `\ **read** <pulp_file-client/ContentFilesApi.md#read>`_
+     - **GET** {file_content_href}
+     - Inspect a file content
+   * - *DistributionsFileApi*
+     - `\ **create** <pulp_file-client/DistributionsFileApi.md#create>`_
```
fabricio-aguiar, line 117: Initially, it breaks here due to 
```
Warning, treated as error:
/home/vagrant/devel/pulp_file/docs/pulp_file-client.rst:3:Duplicate explicit target name: "**create**".

```

docs/pulp_file-client/ContentFilesApi.rst
```diff
@@ -0,0 +1,356 @@
...
+
+Create a file content
+
+Trigger an asynchronous task to create content,optionally create new repository version.
+
+Example
+^^^^^^^
+
+
+* Basic Authentication (Basic):
```
fabricio-aguiar, line 40: then it breaks here with
```
Warning, treated as error:
/home/vagrant/devel/pulp_file/docs/pulp_file-client/ContentFilesApi.rst:40:Inline literal start-string without end-string.
```

docs/pulp_file-client/ContentFilesApi.rst
```diff
@@ -0,0 +1,356 @@
...
+  ```python
```
fabricio-aguiar, line 41: If I remove the * from the previous line it breaks here with:
```
Warning, treated as error:
/home/vagrant/devel/pulp_file/docs/pulp_file-client/ContentFilesApi.rst:41:Inline literal start-string without end-string.
```

fabricio-aguiar:
> Thanks for the PR. How will these docs get updated whenever there is a change to the API? Will it be up to developers to remember to regenerate these docs when they make a change?

@dkliban said it would be manual due to readthedocs, but after trying to build the docs, I suspect `m2r` has some limitations and making these docs compatible with sphinx will need too many hacks

PR #388, opened 2020-04-30
Update travis config

mdellweg:
[noissue]


docs/Makefile
```diff
@@ -42,7 +42,7 @@ clean:
 	-rm -rf $(BUILDDIR)/*
 
 html:
-	curl -o _static/api.json "http://localhost:24817/pulp/api/v3/docs/api.json?plugin=pulp_file"
+	curl -o _static/api.json "http://pulp/pulp/api/v3/docs/api.json?plugin=pulp_file"
```
quba42, line 45: This change makes it impossible to run `make html` from pulplift source development boxes.
Any ideas how to cover both use cases?

quba42: We had a little discussion about it, and came up with the following:
Since we are using a mixture of shell scripts, containers, and Makefiles, it should be possible to parametrize all API calls in a uniform way using environmental variables. This would have the added benefit of reducing the reliance on `sed -i` for parametrization (which is a bit of a hack).

PR #390, opened 2020-05-08
Introducing release script

fao89:
Introducing release script

which executes the first 6 steps of https://pulp.plan.io/projects/pulp/wiki/Pulp3_Release_Guide

https://pulp.plan.io/issues/6600
ref #6600


.travis/release.py
```diff
@@ -0,0 +1,114 @@
...
+            sep = "'" if len(line.split('"')) == 1 else '"'
+            for word in line.split(sep):
+                if "pulpcore" in word:
+                    pulpcore_word = word
+
+            line = line.replace(
+                pulpcore_word, f"pulpcore>={pulpcore_version},<{next_pulpcore_version}"
+            )
+
+        setup_file.write(line)
```
fao89, line 75: Still need to do it, as `bump2version` does not update `pulpcore` requirement

.travis/release.py
```diff
@@ -0,0 +1,114 @@
...
+    for line in setup_lines:
+        if "pulpcore" in line and "entry_points" not in line and "pulpcore" not in release_path:
+            sep = "'" if len(line.split('"')) == 1 else '"'
+            for word in line.split(sep):
+                if "pulpcore" in word:
+                    pulpcore_word = word
+
+            line = line.replace(pulpcore_word, f"pulpcore>={pulpcore_version}")
+
+        setup_file.write(line)
```
fao89, line 99: Still need to do it, as `bump2version` does not update `pulpcore` requirement

.travis/release.py
```diff
@@ -0,0 +1,114 @@
...
+
+from pulp_file import __version__ as current_dev_version
+
+
+release_path = os.path.dirname(os.path.abspath(__file__))
+plugin_path = release_path
+if ".travis" in release_path:
+    plugin_path = os.path.dirname(release_path)
+
+release_part = sys.argv[1]
```
bmbouter, line 38: What do you think about converting this to use [`argparse`](https://docs.python.org/3/library/argparse.html) directly? The benefit is that it provides auto-documenting --help usage on the command which could replace the manually maintained one at the top of this file for example. Also it provides validation.

setup.py
```diff
@@ -7,9 +7,13 @@
 with open("README.rst") as f:
     long_description = f.read()
 
+version = {}
+with open("pulp_file/__init__.py") as fp:
```
bmbouter, line 11: Doesn't bump2verison handle this de-duplication for us? I think this is option (3) listed here but we're also using option (2) from here https://packaging.python.org/guides/single-sourcing-package-version/ Can bump2version be configured to set it directly in both places?

.travis/release.py
```diff
@@ -0,0 +1,157 @@
...
+    formatter_class=argparse.RawTextHelpFormatter, description="Start the release process."
+)
+
+parser.add_argument(
+    "release_part",
+    type=str,
+    help=textwrap.dedent(
+        """\
+                            Whether the release should be major, minor or patch.
+
```
bmbouter, line 53: What is the motivation for making these triple quote strings with dedent?

fao89: I just copied from plugin-template
https://github.com/pulp/plugin_template/blob/master/plugin-template#L58

bmbouter: Is using simple strings an option like the examples in argparse? https://docs.python.org/3/library/argparse.html

setup.py
fao89: I used this approach because when I tried to use `bump2version` pointing to `setup.py` it changed the plugin version and also the requirements version.

Example:
```
requirements = [requests>=3.0.1]
setup(version=3.0.1
```
it would replace 3.0.1 on these 2 places

bmbouter: Oh I see what you mean. This makes sense then.

bmbouter: Not being able to read the version of a package in setup.py itself feels awkward. So not as part of this PR, but maybe in future work the requirements can be moved to its own file in the repository that setup.py loads (similar to the other requirements files we have and then bump2version can handle it all.

.travis/release.py
fao89: it works with simple strings:
```
❯ python .travis/release.py --help 
usage: release.py [-h] release_part lower_pulpcore_version upper_pulpcore_version

Start the release process.

positional arguments:
  release_part          Whether the release should be major, minor or patch.
  lower_pulpcore_version
                        Lower bound of pulpcore requirement.
  upper_pulpcore_version
                        Upper bound of pulpcore requirement.

optional arguments:
  -h, --help            show this help message and exit
```

.travis/release.py
```diff
@@ -0,0 +1,134 @@
+"""
+Release script
+
+$ python .travis/release.py
+
+Args:
```
bmbouter, line 6: With argparse being self-documenting I don't think having a second copy here is more helpful than the risk it will become out of date.

.travis/release.py
```diff
@@ -0,0 +1,134 @@
...
+    release_part - major/minor/patch
+    lower_pulpcore_version - lower bound of pulpcore requirement
+    upper_pulpcore_version - upper bound of pulpcore requirement
+
+Example:
+setup.py on plugin before script:
```
bmbouter, line 12: These examples are maybe good if they were moved into argparse as part of the heltext response also.

.travis/release.py
fao89: agreed, I put it here because I didn't have it documented before as argparse did

setup.py
```diff
@@ -17,7 +18,7 @@
     author_email="pulp-dev@redhat.com",
     url="https://pulpproject.org/",
     python_requires=">=3.6",
-    install_requires=requirements,
+    install_requires=requirements,  # Currently it cannot deal with hashes
```
fao89, line 21: From what I researched, currently, setuptools does not support hashes on `install_requires` yet

hashed_requirements.txt
```diff
@@ -0,0 +1,409 @@
+#
+# This file is autogenerated by pip-compile
+# To update, run:
+#
+#    pip-compile --generate-hashes --output-file=hashed_requirements.txt
+#
+aiodns==2.0.0 \
+    --hash=sha256:815fdef4607474295d68da46978a54481dd1e7be153c7d60f9e72773cd38d77d \
+    --hash=sha256:aaa5ac584f40fe778013df0aa6544bf157799bd3f608364b451840ed2c8688de \
+    # via pulpcore
```
fao89, line 10: I kept this file here because I thought it would be cool to document on the installation steps we have requirements with hashes

fao89:
I created this PoC of moving to poetry:
https://github.com/pulp/pulp_file/pull/391

hashed_requirements.txt
bmbouter: This is great, but how can we maintain it? If it becomes out of date that would be a problem.

fao89: true, I think we have to remove it for now, as it is not supported yet on `install_requires` from `setuptools`

bmbouter: Even though it's sweet, I agree.

setup.py
bmbouter: Since we're not using the hashes I think this comment doesn't make sense anymore.

PR #392, opened 2020-05-19
Speeding up CI with pre-built image

fao89:
[noissue]


pulp_file/tests/performance/pulpperf/interact.py
```diff
@@ -2,10 +2,11 @@
 import requests
 import tempfile
 import time
+import os
 
 from .utils import measureit, urljoin
 
-BASE_ADDR = "http://localhost:24817"
```
fao89, line 8: I believe it was the reason for breaking the performance tests

docs/_scripts/base.sh
```diff
@@ -2,8 +2,8 @@
 set -e
 
 echo "Setting environment variables for default hostname/port for the API and the Content app"
-export BASE_ADDR=${BASE_ADDR:-http://pulp:80}
-export CONTENT_ADDR=${CONTENT_ADDR:-http://pulp:80}
+export BASE_ADDR=${BASE_ADDR:-http://localhost:24817}
+export CONTENT_ADDR=${CONTENT_ADDR:-http://localhost:24816}
```
mikedep333, line 6: Thanks you for restoring this!

We had a user on the ML lately say these scripts didn't work.

PR #398, opened 2020-06-01
Creating a volume for /var/lib/pulp

fao89:
During release CI fails due missing `/var/lib/pulp`
[noissue]

### DIFF
--- a/.travis/install.sh
+++ b/.travis/install.sh
@@ -64,6 +64,7 @@ services:
     image: "pulp:${TAG}"
     volumes:
       - ./settings:/etc/pulp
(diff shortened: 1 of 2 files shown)


fao89:
@dkliban do you think this change would fix this behavior on releases?
https://travis-ci.com/github/pulp/pulp_file/jobs/340696154#L1506

dkliban:
@fao89 We should just create this directory in the Containerfile and not have to worry about creating the volume.

fao89:
I created the directory with the whitenoise PR, so I'm closing this one

PR #406, opened 2020-06-06
Rebase FileContentResource to subclass from BaseContentResource

ggainey:
ref #6807
Required PR: https://github.com/pulp/pulpcore/pull/745

### DIFF
--- a/CHANGES/6936.misc
+++ b/CHANGES/6936.misc
@@ -1,0 +1,1 @@
+Rebased FileContentResource to subclass from BaseContentResource.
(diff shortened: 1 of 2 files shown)


CHANGES/6936.bugfix
```diff
@@ -0,0 +1 @@
+Rebased FileContentResource to subclass from BaseContentResource.
```
daviddavis, line 1: Based on the convo over on the pulpcore PR, this should probably be a misc change too.

ggainey: done

PR #415, opened 2020-07-14
Changes to support associating Remotes with Repositories

daviddavis:
fixes #7135


pulp_file/app/viewsets.py
```diff
@@ -93,7 +93,7 @@ def sync(self, request, pk):
         serializer.is_valid(raise_exception=True)
 
         repository = self.get_object()
-        remote = serializer.validated_data.get("remote")
+        remote = serializer.validated_data.get("remote", repository.remotes.first())
```
bmbouter, line 96: :+1:

PR #420, opened 2020-08-13
Fix name of the artifact field

lubosmj:
closes #5966


CHANGES/5966.docs
```diff
@@ -0,0 +1 @@
+Fixed the name of the artifact field
```
daviddavis, line 1: 5966.*doc*

PR #426, opened 2020-09-08
Require tests for feature/bugfix

fao89:
https://pulp.plan.io/issues/7437
closes #7437

### DIFF
--- a/.travis/pre_before_install.sh
+++ b/.travis/pre_before_install.sh
@@ -1,0 +1,23 @@
+#!/usr/bin/env bash
+
+set -mveuo pipefail
+
+# skip this check for everything but PRs
+if [ "$TRAVIS_PULL_REQUEST" = "false" ]; then
+  exit 0
+fi
+
+if [ "$TRAVIS_COMMIT_RANGE" != "" ]; then
+  RANGE=$TRAVIS_COMMIT_RANGE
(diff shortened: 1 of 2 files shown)


.travis/pre_before_install.sh
```diff
@@ -0,0 +1,30 @@
...
+# Travis sends the ranges with 3 dots. Git only wants two.
+if [[ "$RANGE" == *...* ]]; then
+  RANGE=`echo $TRAVIS_COMMIT_RANGE | sed 's/\.\.\./../'`
+else
+  RANGE="$RANGE~..$RANGE"
+fi
+
+
+REQUIRES_TEST=$(git diff --name-only $RANGE | grep -E 'feature|bugfix')
+CONTAINS_TEST=$(git diff --name-only $RANGE | grep -E '^test_')
```
mdellweg, line 25: I think we should be testing this for each commit in the PR separately (Neccessary requirement for proper cherry-picking). Also i wished, this could deliver a warning instead of a failure, as it is a bit fuzzy by nature... (But that will remain a wish with current CI providers.).
Also I liked the first grep to be specific about files being in "/CHANGES", and maybe the second one expecting a comment like `tests #1234` within the `/pulp_file/tests` directory.

PR #431, opened 2020-09-18
Fix docs publication

fao89:
Fix: https://travis-ci.com/github/pulp/pulp_file/jobs/387174826#L1220
[noissue]


bmbouter (approved):
Thanks!

PR #436, opened 2020-10-05
Update Travis after 1.3.0 release

bmbouter:
* Update `stable_branch` to None
* regenerate the plugin template

[noissue]

### DIFF
--- a/template_config.yml
+++ b/template_config.yml
@@ -44,7 +44,7 @@ pulpprojectdotorg_key_id: aa499d7938ed
 pydocstyle: true
 pypi_username: pulp
 redmine_project: pulp_file
-stable_branch: '1.2'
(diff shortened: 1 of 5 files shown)


template_config.yml
```diff
@@ -44,7 +44,7 @@ pulpprojectdotorg_key_id: aa499d7938ed
 pydocstyle: true
 pypi_username: pulp
 redmine_project: pulp_file
-stable_branch: '1.2'
+stable_branch: None
```
daviddavis, line 47: I think "null" would be better here since it's valid yaml.

daviddavis (approved):

daviddavis (approved):
One small suggestion. Approving.

PR #440, opened 2020-11-12
Add some basic github action lints

dralley:
Not suggesting that we start migrating yet, but it'd be nice to have a little bit of "insurance policy" ready just in case they do decide to cut us off before responding to our emails.  If that happens we could just copy this to the other plugins and stick to running the tests manually temporarily.

Or we could implement this small bit of work in a couple of different CI services and compare them.  In any case it didn't take all that long to do (thanks to @fao89 's work earlier this year) so it didn't waste that much time.

[noissue]


mdellweg:
How about collecting the agnostic scripts in a directory called ".ci" instead of ".travis" or ".github".
I think it would give us a measure to see how far we are from migrating.
And maybe it will bring us closer to "i can run all the CI checks locally".

dralley:
Closing.  We'll do NPM first and then update this one from the plugin template.

PR #451, opened 2020-12-08
Moves version to PulpPluginAppConfig

bmbouter:
This redeploys the plugin template and updates a few files that should
be updated to match but are not plugin template managed. Specifically:

* Removing `__version__` from `pulp_file/__init__.py`
* Adding the version string directly to `docs/conf.py`
* Adding the `version` attribute to `PulpPluginAppConfig` subclass in
  `pulp_file/app/__init__.py`.
* Updating .bumpversion.cfg

[noissue]


daviddavis (approved):

.bumpversion.cfg
```diff
@@ -14,6 +14,8 @@ values =
 	dev
 	prod
 
-[bumpversion:file:./pulp_file/__init__.py]
+[bumpversion:file:./pulp_ansible/app/__init__.py]
```
dkliban, line 17: you mean `./pulp_file/app` probably

bmbouter: the copypasta!

PR #456, opened 2020-12-17
Release 1.5.0 

dkliban:
This PR contains 4 commits. The first one modifies how functional tests are run to support cases where functional tests on master branch of pulpcore are no longer compatible with the pulpcore release on PyPI. The second commit downgrades pulp-smash if the pulpcore version starts with 3.9. This commit should be reverted once pulpcore 3.10 is available on PyPI. The other 2 commits deal with with version strings.


.github/workflows/scripts/func_test_script.sh
```diff
@@ -4,4 +4,16 @@
...
+
+if [ "${GITHUB_REF##refs/tags/}" != "${GITHUB_REF}" ]
+then
+  PULPCORE_VERSION=$(http http://pulp/pulp/api/v3/status/ | jq -r '.versions'[0].version)
+  cd ../pulpcore
+  git fetch origin refs/tags/${PULPCORE_VERSION}
+  git checkout FETCH_HEAD
+  cd ../pulp_file
+  # Temporarily need to downgrade pulp-smash to run pulpcore 3.9 tests
+  pip install 'pulp-smash==1!0.12.0'
```
mdellweg, line 16: shouldn't this be taken from the `functest_requirements.txt` of pulpcore?
If that does not work you might note when to remove this workaround.

mdellweg (approved):

mdellweg (approved):

PR #461, opened 2021-01-29
Update the port for pulp_fixtures

daviddavis:
[noissue]


mdellweg (approved):

mdellweg (approved):
Assuming this is done by the virtue of the plugin_template.

PR #463, opened 2021-02-01
Update CI files from plugin_template

daviddavis:
[noissue]


.github/workflows/scripts/script.sh
```diff
@@ -31,6 +31,11 @@ if [[ "$TEST" = "docs" || "$TEST" = "publish" ]]; then
   make PULP_URL="http://pulp" html
   cd ..
 
+  echo "Validating OpenAPI schema..."
+  cat $PWD/.ci/scripts/schema.py | cmd_stdin_prefix bash -c "cat > /tmp/schema.py"
+  cmd_prefix bash -c "python3 /tmp/schema.py"
+  # cmd_prefix bash -c "pulpcore-manager spectacular --file pulp_schema.yml --validate"
```
daviddavis, line 37: @fao89 are the schema parts of this PR correct? It seems like this code isn't doing anything?

fao89: yes, it is because the fix for OpenAPI schema is only available on pulpcore 3.10, during open floor we decided to disable it till pulpcore 3.10 is released, otherwise, it could break plugin releases

fao89 (approved):

mdellweg (approved):

PR #464, opened 2021-02-05
Minor refactor of Manifest class

dralley:
[noissue]

This mirrors the way that ```Entry``` class works.

### DIFF
--- a/pulp_file/app/tasks/synchronizing.py
+++ b/pulp_file/app/tasks/synchronizing.py
@@ -75,10 +75,12 @@ class FileFirstStage(Stage):
 
         with ProgressReport(message="Parsing Metadata Lines", code="parsing.metadata") as pb:
             manifest = Manifest(result.path)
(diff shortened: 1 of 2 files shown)


pulp_file/app/tasks/synchronizing.py
```diff
@@ -75,10 +75,12 @@ async def run(self):
 
         with ProgressReport(message="Parsing Metadata Lines", code="parsing.metadata") as pb:
             manifest = Manifest(result.path)
-            pb.total = manifest.count()
```
dralley, line 78: We're opening the file twice, parsing it twice, and iterating it twice because of ```.count()```.  It'd be better to just open and parse it once, and re-use the same list of entries.

Even with tens of thousands of files the list shouldn't be massive.  Our perf repo with 20k files has a manifest around 1.5 mb.

daviddavis (approved):

PR #465, opened 2021-02-12
Remove unused CONTENT_ADDR variable from docs test

daviddavis:
We're using the distribution base url now.

[noissue]


mdellweg (approved):

mdellweg:
Is this a plugin template kind of change?

daviddavis:
No, the plugin_template only provides a hook in the CI to call this file if it exists and there's nothing in the CI that references this file.

PR #466, opened 2021-02-12
Enable CLI test for pulp_file

daviddavis:
[noissue]


.github/workflows/scripts/script.sh
```diff
@@ -111,6 +111,9 @@ if [ -f $FUNC_TEST_SCRIPT ]; then
 else
     pytest -v -r sx --color=yes --pyargs pulp_file.tests.functional
 fi
+pushd ../pulp-cli
+pytest -m pulp_file
```
mdellweg, line 115:
```suggestion
pytest -v -m pulp_file
```

mdellweg (approved):
ACK, however it looks like this is a change actually happening in plugin_template.

PR #467, opened 2021-02-15
Add support for automatic publishing and distributing

dralley:
closes: #7469
https://pulp.plan.io/issues/7469


daviddavis:
You have a typo--should be `closes: #7469`.

dralley:
test failure is just caused by the requirement of pulpcore 3.11 which doesn't exist yet.  and I suppose it doesn't think the prerelease counts

mdellweg:
Make it depend on 3.11.dev and it should be fine

pulp_file/app/serializers.py
```diff
@@ -99,6 +129,7 @@ class FilePublicationSerializer(PublicationSerializer):
     manifest = serializers.CharField(
```
dralley, line 129: Should I remove the fields from the Publication model and proxy them through PublishSettings?

pulp_file/app/tasks/publishing.py
```diff
@@ -7,23 +7,25 @@
...
-from pulp_file.app.models import FileContent, FilePublication
+from pulp_file.app.models import FileContent, FilePublication, FilePublishSettings
 from pulp_file.manifest import Entry, Manifest
 
 
 log = logging.getLogger(__name__)
 
 
-def publish(manifest, repository_version_pk):
+def publish(repository_version_pk, publish_settings_pk):
```
dralley, line 17: Ugh, I might have to rethink how to structure this to accomodate the migration plugin.  Which uses ```publish()``` directly.

pulp_file/app/serializers.py
daviddavis: That's a good question. I lean towards yes (to remove the duplicate data) but I don't have a strong opinion.

daviddavis:
I tested this out and it worked great :100: 

A couple things: 
1. How hard would it be to add progress reports for publishing and distributing to the task?
2. We could really use some docs (or something I could point the students who are working on the UI to)

dralley:
Not hard.  There's not much to report, but we can just add a report that says it did a publish.

pulp_file/app/migrations/0009_auto_publish.py
```diff
@@ -0,0 +1,43 @@
...
+from django.db import migrations, models
+import django.db.models.deletion
+
+
+def populate_publish_settings(apps, schema_editor):
+    FilePublication = apps.get_model('file', 'filepublication')
+    FilePublishSettings = apps.get_model('file', 'filepublishsettings')
+
+    for publication in FilePublication.objects.iterator():
+        publish_settings, _ = FilePublishSettings.objects.get_or_create(
```
mdellweg, line 12: I know, this is the recommended way to ignore partial return values. But does it work together with _ binding to gettext?

dralley: In this particular case, there is no gettext import for it to conflict with, but in the general case, I don't know.

pulp_file/app/migrations/0010_migrate_abstract_fields.py
```diff
@@ -0,0 +1,31 @@
+# Generated by Django 2.2.19 on 2021-02-23 18:05
+
+from django.db import migrations
+
+
+def migrate_distribution_fields(apps, schema_editor):
+    # Set the core "_publication" field with the value of the detail "publication" field
+    FileDistribution = apps.get_model('file', 'filedistribution')
+
+    distributions = list(FileDistribution.objects.all())
```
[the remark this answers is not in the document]
dralley: It would be really difficult to have enough distributions for that to be a problem (you'd need hundreds of thousands of them) but maybe I can write this better.

pulp_file/app/migrations/0010_migrate_abstract_fields.py
```diff
@@ -0,0 +1,31 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('file', '0009_auto_publish'),
+        ('core', '0060_add_new_distribution_fields'),
+    ]
+
+    operations = [
+        migrations.RunPython(migrate_distribution_fields),
```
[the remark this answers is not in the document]
dralley: It looks like none of our other migrations provide down migrations, and this one would be particularly difficult since the core tables are involved also.

pulp_file/app/migrations/0009_auto_publish.py
mdellweg: I agree, gettext would make the migrations really unstable.

PR #469, opened 2021-02-23
[DO NOT MERGE] Use fixtures.pulpproject.org

daviddavis:
[noissue]


daviddavis:
@mscherer this PR runs against fixtures.pulpproject.org. It should tell you whether the site is working or not.

PR #482, opened 2021-03-25
Switches `FileDistribution` use `Distribution`

bmbouter:
pulpcore==3.12 introduces a new MasterModel named `Distribution`
designed to replace the `BaseDistribution` MasterModel. This PR switches
the `FileDistribution` to use `Distribution`. It also ships a migration
which moves the data from the `BaseDistribution` table to the
`Distribution` field.

Required PR: https://github.com/pulp/pulpcore/pull/1198/

closes #8387


bmbouter:
Aaaand it happened again. Apparently where I put tmate in the workflow causes GHA to lock github. :open_mouth:

PR #483, opened 2021-03-25
Make filesystem exporter apis async

mdellweg:
This will introduce a change that is about to come with pulpcore 3.12
anyway. But that would introduce incompatible bindings changes between
two pulp_file releases. We must be proactive here in order to be able
to ship compatible binding packages.

fixes #8451
https://pulp.plan.io/issues/8451


mdellweg:
This is failing in a pulpcore test now. I don't know how to get that test green from the pulp_file side.

daviddavis:
You can use the "Required PR" feature to include a pulpcore PR (see [here](https://git.io/JYtUs)). Does that not work?

CHANGES/8451.bugfix
```diff
@@ -0,0 +1 @@
+Added asynchronous tasking to the Update and Delete endpoints of FilesystemExporter to provide proper locking on resources.
```
daviddavis, line 1: I might also add a 8451.removal as well saying that these endpoints return tasks/202. But this may be overkill.

pulp_file/app/viewsets.py
```diff
@@ -194,6 +195,60 @@ class FileFilesystemExporterViewSet(ExporterViewSet):
     queryset = FileFilesystemExporter.objects.all()
     serializer_class = FileFilesystemExporterSerializer
 
+    # ----8<----8<----8<----
+    # This section can be savely removed, once the plugin depends on pulpcore >= 3.12
```
daviddavis, line 199: Is there a task filed for this?

daviddavis (approved):

pulp_file/app/viewsets.py
mdellweg: https://pulp.plan.io/issues/8465

mdellweg: It needs to go into the 1.8.0 milestone i think.

PR #492, opened 2021-04-02
Add support for automatic publishing and distributing

dralley:
closes: #7469
https://pulp.plan.io/issues/7469


dralley:
See note: https://github.com/pulp/pulpcore/pull/1229#issuecomment-812282405

dralley:
Without `PublishSettings`, these PRs are pretty tiny.

pulp_file/app/models.py
```diff
@@ -60,9 +60,37 @@ class FileRepository(Repository):
...
+                manifest=self.manifest,
+                repository_version_pk=version.pk,
+            )
+
+            distributions = self.distributions.all()
+
+            if publication and distributions:
+                for distribution in distributions:
+                    distribution.publication = publication
+                    distribution.save()
```
mdellweg, line 92: My impression is, that this will update _all_ distributions that are pointing to any repoversion of that repository.
I think we need a way to distinguish distributions that are meant to stay what they are from ones that are meant to "always server a publication of the latest version".

ipanova: we could filter out distributions that point to a repo_version and instead update only those that point to a repo

pulp_file/app/serializers.py
```diff
@@ -62,8 +62,23 @@ class FileRepositorySerializer(RepositorySerializer):
...
+        help_text=_(
+            "Whether to automatically create publications for new repository versions, "
+            "and update any distributions pointing to this repository."
+        ),
+        default=False,
+        required=False,
+    )
+
+    manifest = serializers.CharField(
+        help_text=_("Filename to use for manifest file containing metadata for all the files."),
```
daviddavis, line 75: Is this field used just for autopublish? Or does it also apply to manual publishes?

daviddavis:
I'd suggest adding an example of sync+publish to https://docs.pulpproject.org/pulp_file/workflows/sync.html

Also, I wonder if we should label this as a tech preview?

pulp_file/app/serializers.py
dralley: Right now, only autopublish.  It's an optional parameter at publication time (when creating publications manually) but if you don't supply it it will fall back to the serializer default rather than the repository default.

dralley: @daviddavis Although, it may not be so difficult to make it use the repository default. Should I go ahead and try?

dralley: The new commit does this.  Yes or no?

daviddavis: Eh, we can add it later if we decide to.

dralley:
@daviddavis I realized that we need to get the CLI updated with the new commands before this documentation would actually work - therefore - it might be better handled in a separate issue.  What do you think?

How much further into detail should it go beyond what is written here:

```rst
Automate Publication and Distribution
-------------------------------------

With a little more initial setup, you can have publications and distributions for your repositories
updated automatically when new repository versions are created.

.. code-block:: bash

    # This configures the repository to produce new publications when a new version is created
    pulp file repository update --name $REPO_NAME --autopublish True

    # This configures the distribution to be track the latest repository version for a given repository
    pulp file distribution update --name $DIST_NAME --repository $REPO_NAME

```

pulp_file/app/models.py
mdellweg: Reading it again, i may have misunderstood. This is only updating the distributions, that explicitely point to a repository, and are thereby setup to be in autopublish mode.
Distributions, that are created by pointing to a repo_version only will not be touched.

daviddavis:
Separate issue is fine. The docs look good :+1:.

daviddavis (approved):

pulp_file/app/models.py
ipanova: you are right

dralley: Yes.  One ambiguity is that if you create a Distribution pointing to one repository version and then attach a totally separate repository, at some point in the future the Distribution will get overwritten a the new repository version.  So there's possibly an argument to be made for being more strict about what we allow users to do.

PR #493, opened 2021-04-05
Improve publish speed by between 2x and 20x

dralley:
Use a more single query rather than N small queries to iterate content
artifacts.

Using a repositoriy containing 20,000 content units, the publish time
is improved to around 1 second (opposed to ~23 seconds) if the content
is immediate, and 14 seconds (vs. the same) if content is on_demand.

Complexity (in terms of # of queries performed and not taking into account
query complexity) was reduced from 2N + 1 to either a constant or 1N
depending on whether content was synced immediate or on-demand.

Benefit is very noticeable.

closes: #8508
https://pulp.plan.io/issues/8508

### DIFF
--- a/pulp_file/app/tasks/publishing.py
+++ b/pulp_file/app/tasks/publishing.py
@@ -5,9 +5,14 @@ from gettext import gettext as _
 
 from django.core.files import File
 
-from pulpcore.plugin.models import RepositoryVersion, PublishedMetadata, RemoteArtifact
+from pulpcore.plugin.models import (
+    ContentArtifact,
+    RepositoryVersion,
+    PublishedMetadata,
+    RemoteArtifact,
+)
 
-from pulp_file.app.models import FileContent, FilePublication
+from pulp_file.app.models import FilePublication
 from pulp_file.manifest import Entry, Manifest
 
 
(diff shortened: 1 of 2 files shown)


dralley:
Complexity (in terms of # of queries performed and not taking into account query complexity) 2N + 1, and is now constant or 1N depending on whether content was synced immediate or on-demand.

Benefit is very noticeable.

mdellweg (approved):
:+1:

If i read it correctly, it scales accordingly for mixed repo versions: `1 + #(on-demand artifacts)`

ipanova:
Nice!

pulp_file/app/tasks/publishing.py
```diff
@@ -57,20 +62,19 @@ def populate(publication):
...
-        _artifact = content_artifact.artifact
-        if not _artifact:
-            _artifact = RemoteArtifact.objects.filter(content_artifact=content_artifact).first()
-        return _artifact
-
-    for content in FileContent.objects.filter(
-        pk__in=publication.repository_version.content
-    ).order_by("-pulp_created"):
-        for content_artifact in content.contentartifact_set.all():
-            artifact = find_artifact()
```
dralley, line 70: 2N + 1 queries

PR #495, opened 2021-04-06
Add a reverse migration to distribution change

mdellweg:


mdellweg:
It turns out, that this operation fails to perform backwards. I have no idea yet, how to solve this.

``` python
        migrations.AlterField(
            model_name='filedistribution',
            name='distribution_ptr',
            field=models.OneToOneField(auto_created=True,
                                       on_delete=django.db.models.deletion.CASCADE,
                                       parent_link=True, primary_key=True,
                                       related_name='file_filedistribution', serialize=False,
                                       to='core.Distribution'),
            preserve_default=False,
        ),
```

mdellweg:
It turns out, that the following statements in the resulting sql-backward migration (pulpcore-manager sqlmigrate file 0009 --backwards) are in the wrong order:
``` sql
ALTER TABLE "file_filedistribution" ALTER COLUMN "distribution_ptr_id" DROP NOT NULL;
ALTER TABLE "file_filedistribution" DROP CONSTRAINT "file_filedistribution_distribution_ptr_id_5ce5564b_pk";
```
I think, this is a django bug.

daviddavis:
This looks good to me.

mdellweg:
> This looks good to me.

It does not work. Once loaded with data, you get:
django.db.utils.InternalError: could not find trigger 86175

mdellweg:
This should work now. I tested forward and backward several times with the following data:

```
#!/bin/bash

set -eux

pulp file distribution destroy --name "test2" || true
pulp file distribution destroy --name "test1" || true
pulp file repository destroy --name "test1" || true
pulp file remote destroy --name "test1" || true

pulp file remote create --name "test1" --url "https://fixtures.pulpproject.org/file/PULP_MANIFEST"
pulp file repository create --name "test1" --remote "test1"
pulp file repository sync --name "test1"
PUBLICATION_HREF=$(pulp file publication create --repository "test1" | jq -r '.pulp_href')
pulp file distribution create --name "test1" --base-path "test1"
pulp file distribution create --name "test2" --base-path "test2" --publication $PUBLICATION_HREF
```

daviddavis:
I created some distributions against `1c96911` then checked out this PR and ran the migration. Everything worked. So then I ran `pulpcore-manager migrate file 0008` followed by `pulpcore-manager migrate file 0009` and confirmed everything worked again. :+1:

daviddavis (approved):

daviddavis:
I guess we need to update this @pulp/rpm migration too?

https://github.com/pulp/pulp_rpm/blob/master/pulp_rpm/app/migrations/0032_new_distribution_model.py

dralley:
@daviddavis Yup

pulp_file/app/migrations/0009_move_data_to_new_master_distribution_model.py
```diff
@@ -4,73 +4,74 @@
...
-pks_to_delete = []
-
-
-def migrate_data_from_old_master_model_to_new_master_model(apps, schema_editor):
+def migrate_data_from_old_model_to_new_model_up(apps, schema_editor):
+    """ Move objects from FileDistribution to NewFileDistribution."""
     FileDistribution = apps.get_model('file', 'FileDistribution')
-    CoreDistribution = apps.get_model('core', 'Distribution')
-    for old_file_distribution in FileDistribution.objects.all():
+    NewFileDistribution = apps.get_model('file', 'NewFileDistribution')
```
dralley, line 10: Is this just an attempt at making it more generic and proper or is it strictly necessary for functional reasons for the file plugin?

pulp_file/app/migrations/0009_move_data_to_new_master_distribution_model.py
```diff
@@ -4,73 +4,74 @@
...
-                name=old_file_distribution.name,
-                base_path=old_file_distribution.base_path,
-                content_guard=old_file_distribution.content_guard,
-                remote=old_file_distribution.remote,
-                publication=old_file_distribution.publication,
-            )
-            new_master_model_entry.save()
-            old_file_distribution.distribution_ptr = new_master_model_entry
-            old_file_distribution.save()
-            pks_to_delete.append(old_file_distribution.pulp_id)
```
dralley, line 29: I'm glad we can get rid of this, I'm just concerned that perhaps there was some necessary reason it was done in such a convoluted fashion originally?

pulp_file/app/migrations/0009_move_data_to_new_master_distribution_model.py
```diff
@@ -4,73 +4,74 @@
...
+                base_path=file_distribution.base_path,
+                content_guard=file_distribution.content_guard,
+                remote=file_distribution.remote,
+                publication=file_distribution.publication
+            ).save()
+            file_distribution.delete()
 
 
 class Migration(migrations.Migration):
+    atomic = False
```
dralley, line 48: Why is this necessary?

pulp_file/app/migrations/0009_move_data_to_new_master_distribution_model.py
mdellweg: Not sure, what you mean. The `apps.get_model` is Django's way to provide you the historical model with a schema that matches the current state of the migration.
In this case, it provides you the transient `NewFileDistribution` that only pops into existance to later be renamed to `FileDistribution`. Its evanescent nature manifests in its complete lack of presence in the actual code.

pulp_file/app/migrations/0009_move_data_to_new_master_distribution_model.py
mdellweg: I believe the problem of being unable to delete the old base objects was, that the `basedistribution_base_ptr` had a foreign key constraint that was only lifted by removing the field from the table.
In the new approach there is no such ambiguity of inheritance that prevents the immediate deletion of old objects.

pulp_file/app/migrations/0009_move_data_to_new_master_distribution_model.py
mdellweg: This solves the aforementioned trigger error.
My theory: When inserting new items into the new table, Django or Postgres add a trigger to check for constraints upon leaving the transaction. Renaming the table however breaks the relation and the trigger cannot be found anymore. Ending the transaction earlier fixes the problem.
Django suggests making data migrations non atomic:
https://docs.djangoproject.com/en/2.2/howto/writing-migrations/#non-atomic-migrations

pulp_file/app/migrations/0009_move_data_to_new_master_distribution_model.py
dralley: What I mean is that NewFileDistribution and core.Distribution are effectively the same synce the former adds no new fields over the latter.  I agree that it should be used since it's more specific, I'm just trying to make sure my mental model is correct.

pulp_file/app/migrations/0009_move_data_to_new_master_distribution_model.py
dralley: +1

pulp_file/app/migrations/0009_move_data_to_new_master_distribution_model.py
dralley: I know that constraints are checked last, I don't know the specific implementation details.  But this sounds plausible. It also sounds potentially like a bug - maybe we should bring it up in #django and what people say.

Are there portions of the migration that we can make atomic or non-atomic selectively rather than making the entire migration non-atomic in a blanket way?

pulp_file/app/migrations/0009_move_data_to_new_master_distribution_model.py
```diff
@@ -4,73 +4,74 @@
...
+            options={
+                'default_related_name': '%(app_label)s_%(model_name)s',
+            },
+            bases=('core.distribution',),
         ),
-        migrations.RunPython(migrate_data_from_old_master_model_to_new_master_model),
-        migrations.RemoveField(
-            model_name='filedistribution',
-            name='basedistribution_ptr',
+        migrations.RunPython(
```
dralley, line 66: What I mean is, I believe we can put `atomic=True` here rather than having our own transactions inside, and perhaps for other operations also.

Sidenote: if this is declared atomic by default, it means that the "transaction" in the migration was actually just a "savepoint".  I don't remember how the semantic differ, precisely.

pulp_file/app/migrations/0009_move_data_to_new_master_distribution_model.py
mdellweg: NewFileDistribution here is more than core.Distribution, as it's a detail table. So i am copying objects from one detail table to the other.
IMHO you should never add objects to the core.Distribution table directly.

pulp_file/app/migrations/0009_move_data_to_new_master_distribution_model.py
ggainey: transactions in postgres can have multiple savepoints, and rollback can go back to a specific savepoint (rather than rolling back the entire transaction). You have to specifically do a 'ROLLBACK TO savepoint', iirc, to take advantage of this, and I don't know if/how Django does so.

pulp_file/app/migrations/0009_move_data_to_new_master_distribution_model.py
mdellweg: I have not seen a way to make several steps of the migration atomic.
What we could do is move the last two steps (Deleting and renaming Tables) in a separate migration. But i am unsure whether this would gain much.

pulp_file/app/migrations/0009_move_data_to_new_master_distribution_model.py
mdellweg: I had atomic=true with and without the inner transactions and it produced the same error.

dralley: Good to know

dralley: I'm still interested in figuring out of this is an upstream bug, but LGTM

dralley (approved):

dralley:
@mdellweg I'm planning on swapping the distribution models today on many of the plugins.  If you can copy this PR to the RPM plugin and fix the template to use the new model, then I'll handle all of the rest.

mdellweg:
> @mdellweg I'm planning on swapping the distribution models today on many of the plugins. If you can copy this PR to the RPM plugin and fix the template to use the new model, then I'll handle all of the rest.

What template do you mean?

dralley:
@mdellweg The plugin template. e.g. https://github.com/pulp/plugin_template/blob/master/templates/bootstrap/plugin_name/app/viewsets.py.j2#L218

daviddavis:
Oh yea, I totally forgot about the plugin template. I can take care of it unless @mdellweg you want to.

mdellweg:
@dralley https://github.com/pulp/pulp_rpm/pull/1971

@daviddavis that would be great!

PR #502, opened 2021-05-01
Sync with metadata mirroring

dralley:
[noissue]

### DIFF
--- a/pulp_file/app/tasks/synchronizing.py
+++ b/pulp_file/app/tasks/synchronizing.py
@@ -4,7 +4,9 @@ import os
 from gettext import gettext as _
 from urllib.parse import urlparse, urlunparse
 
-from pulpcore.plugin.models import Artifact, ProgressReport, Remote
+from django.core.files import File
+
+from pulpcore.plugin.models import Artifact, ProgressReport, Remote, PublishedMetadata
 from pulpcore.plugin.stages import (
     DeclarativeArtifact,
     DeclarativeContent,
@@ -12,11 +14,14 @@ from pulpcore.plugin.stages import (
     Stage,
 )
 
-from pulp_file.app.models import FileContent, FileRemote, FileRepository
+from pulp_file.app.models import FileContent, FileRemote, FileRepository, FilePublication
 from pulp_file.manifest import Manifest
 
 
 log = logging.getLogger(__name__)
+
+
+metadata_files = []
 
 
 def synchronize(remote_pk, repository_pk, mirror):
@@ -42,7 +47,24 @@ def synchronize(remote_pk, repository_pk, mirror):
 
     first_stage = FileFirstStage(remote)
     dv = DeclarativeVersion(first_stage, repository, mirror=mirror)
-    return dv.create()
+    rv = dv.create()
+    if mirror:
+        # TODO: this is awful, we really should rewrite the DeclarativeVersion API to
+        # accomodate this use case
+        global metadata_files
+        with FilePublication.create(rv, pass_through=True) as publication:
+            (mdfile_path, relative_path) = metadata_files.pop()
+            PublishedMetadata.create_from_file(
+                file=File(open(mdfile_path, "rb")),
+                relative_path=relative_path,
+                publication=publication,
+            )
+            publication.manifest = relative_path
+            publication.save()
+
+        log.info(_("Publication: {publication} created").format(publication=publication.pk))
+
+    return rv
 
 
 class FileFirstStage(Stage):
@@ -65,6 +87,8 @@ class FileFirstStage(Stage):
         """
         Build and emit `DeclarativeContent` from the Manifest data.
         """
+        global metadata_files
+
         deferred_download = self.remote.policy != Remote.IMMEDIATE  # Interpret download policy
         with ProgressReport(message="Downloading Metadata", code="sync.downloading.metadata") as pb:
             parsed_url = urlparse(self.remote.url)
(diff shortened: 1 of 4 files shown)


pulp_file/app/tasks/synchronizing.py
```diff
@@ -42,7 +47,24 @@ def synchronize(remote_pk, repository_pk, mirror):
 
     first_stage = FileFirstStage(remote)
     dv = DeclarativeVersion(first_stage, repository, mirror=mirror)
-    return dv.create()
+    rv = dv.create()
+    if mirror:
+        # TODO: this is awful, we really should rewrite the DeclarativeVersion API to
+        # accomodate this use case
+        global metadata_files
```
dralley, line 54: Yeah this is not ideal.  I made a solid attempt at the `DeclarativeVersion` changes but they are not trivial - it would require a lot of planning.

mdellweg: Is PublishedMetadata a kind of content? In that case we could just add it to the RV as everything else.
We'd need to exempt it from normal publishing however.

dralley: It is, technically, but not "really".  The way it is now is kind of awful, actually.  https://pulp.plan.io/issues/8481

dralley:
I guess the question is, do we want to merge this as-is (with some tests), and then fix it later, or wait until we can add support to DeclarativeVersion, which might take a few cycles especially if the changes are invasive.

pulp_file/app/tasks/synchronizing.py
```diff
@@ -42,7 +47,24 @@ def synchronize(remote_pk, repository_pk, mirror):
 
     first_stage = FileFirstStage(remote)
     dv = DeclarativeVersion(first_stage, repository, mirror=mirror)
-    return dv.create()
+    rv = dv.create()
+    if mirror:
```
mdellweg, line 51: Wasn't the `mirror` variable meant to be a different thing? i.e. distiguishing additive sync from the other one?

dralley: That is what it's for, but the use case here is basically the same.  Before we were saying "give me an exact copy of the remote repository - remove any content from the repository that isn't present on the remote to do so", and now we're extending that to mean "and since Pulp's published metadata should be identical, go ahead and make a publication for this repository using the remote metadata".

pulp_file/app/tasks/synchronizing.py
```diff
@@ -42,7 +47,24 @@ def synchronize(remote_pk, repository_pk, mirror):
...
+        # TODO: this is awful, we really should rewrite the DeclarativeVersion API to
+        # accomodate this use case
+        global metadata_files
+        with FilePublication.create(rv, pass_through=True) as publication:
+            (mdfile_path, relative_path) = metadata_files.pop()
```
mdellweg, line 56: This looks strange to me: `metadata_files` is a list, and we only pop a single entry from it. Why don't we just put that tuple in the variable? Or if we want to make it more portable we should loop over that list.

dralley: I was just trying to keep it consistent with the RPM plugin version I'm working on - there's no reason we couldn't do that.  You're right about looping over the list though.

dralley: Since then I've made changes that mess up the "consistency", anyway.

ipanova:
> I guess the question is, do we want to merge this as-is (with some tests), and then fix it later, or wait until we can add support to DeclarativeVersion, which might take a few cycles especially if the changes are invasive.
> 
> I kind of think we need to do it now, especially since the RPM plugin has no choice.

The main driver of this feature is RPM plugin which also has tight deadlines with regards to the delivery dates.
Based on this ticket there is not seem to be an easy solution to adapt these changes an incorporate into DV https://pulp.plan.io/issues/8687.

My suggestions would be to take the approach we have as of today and work towards a better one later.

dralley:
The RPM version of this PR has now merged

pulp_file/tests/functional/api/test_sync.py
```diff
@@ -82,6 +84,46 @@ def test_sync(self):
...
+        self.repo = self.repo_api.create(gen_repo())
+
+    def tearDown(self):
+        """Clean up."""
+        monitor_task(self.repo_api.delete(self.repo.pulp_href).task)
+        monitor_task(self.remote_api.delete(self.remote.pulp_href).task)
+
+    def test_01_sync(self):
+        """Assert that syncing the repository w/ mirror=True creates a publication."""
+        self.assertEqual(self.publications_api.list().count, 0)
```
mdellweg, line 113: This assumption prevents to run this test in a box you are doing other experiments.
We should instead look if the number increased by one.

dralley: Are we not making that assumption in a great many places?  Any test that cleans up orphan content afterwards for example.

I don't think Pytest does any parallelization.

mdellweg: publications are not cleaned by orphan cleanup. So that doesn't apply here.
This would really require a pclean.

dralley: The problem is that the list of created resources isn't separated by type (like e.g. the repository version summary), so trying to verify this from the created resources is kind of hacky (but doable if it's strongly desired).

dralley: Let me know if this looks better.

mdellweg: Sorry, i should have explained better: I was thinking about assigning the number of publications to a variable without asserting before and assert on `length==N+1` after the autopublish.
But the way it is now looks even better.
BTW, I have seen codepaths to rely on the order of `created_resources`. As in `version = c_r[0]; publication = c_r[1]`.

mdellweg (approved):

dralley:
I've created an issue for this feature

PR #505, opened 2021-05-17
Fix distribution base url on docs tests

fao89:
I suspect double slashes are breaking pulp-operator test:
https://github.com/pulp/pulp-operator/runs/2601467510?check_suite_focus=true#step:11:799

### DIFF
--- a/.github/workflows/scripts/before_install.sh
+++ b/.github/workflows/scripts/before_install.sh
@@ -86,20 +86,6 @@ sed -i -e 's/:24817/pulp/g' generate.sh
 cd ..
 
 
-git clone --depth=1 https://github.com/pulp/pulp-cli.git
-if [ -n "$PULP_CLI_PR_NUMBER" ]; then
-  cd pulp-cli
-  git fetch origin pull/$PULP_CLI_PR_NUMBER/head:$PULP_CLI_PR_NUMBER
-  git checkout $PULP_CLI_PR_NUMBER
-  cd ..
-fi
-
-cd pulp-cli
-pip install -e .
-pulp config create --base-url http://pulp --location tests/settings.toml
-cd ..
-
-
 git clone --depth=1 https://github.com/pulp/pulpcore.git --branch master
 
 cd pulpcore
@@ -122,6 +108,18 @@ fi
 
 pip install --upgrade --force-reinstall ./pulp-smash
 
+git clone --depth=1 https://github.com/pulp/pulp-cli.git
+if [ -n "$PULP_CLI_PR_NUMBER" ]; then
+  cd pulp-cli
+  git fetch origin pull/$PULP_CLI_PR_NUMBER/head:$PULP_CLI_PR_NUMBER
+  git checkout $PULP_CLI_PR_NUMBER
+  cd ..
+fi
+
+cd pulp-cli
+pip install -e .
+pulp config create --base-url http://pulp --location tests/settings.toml
+cd ..
 
 git clone --depth=1 https://github.com/pulp/pulp-certguard.git --branch master
 if [ -n "$PULP_CERTGUARD_PR_NUMBER" ]; then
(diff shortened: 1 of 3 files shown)


dkliban (approved):

.github/workflows/ci.yml
```diff
@@ -114,7 +114,9 @@ jobs:
           SECRETS_CONTEXT: ${{ toJson(secrets) }}
 
       - name: Script
-        run: .github/workflows/scripts/script.sh
+        run: |
+          pip install "click<8.0"
```
bmbouter, line 118: The new RQ==1.8.1 release is out which is compatible with click 8.0. https://pypi.org/project/rq/#history

Do we still need to restrict this?

fao89: yes, CI was breaking before
https://github.com/pulp/pulp_file/runs/2601995311?check_suite_focus=true#step:9:1397
it seems pulp-cli still incompatible with 8.0

fao89: I put it here on purpose, when applying CI template it will be removed, so it will be undone naturally.
I have no idea about pulp-cli release timeline

mdellweg: I think, @daviddavis wants to release pulp-cli 0.9.0 with a pinned version of click today.

daviddavis: Once this job is done, pulp-cli 0.9.0 should be on pypi: https://github.com/pulp/pulp-cli/actions/runs/850630787

fao89: It still breaking, because pulp-smash is forcing click 8.0.0
```
Attempting uninstall: click
    Found existing installation: click 7.1.2
    Uninstalling click-7.1.2:
      Successfully uninstalled click-7.1.2
```
https://github.com/pulp/pulp_file/pull/505/checks?check_run_id=2604129145#step:5:247

fao89: probably not the best solution but I think this would fix:
https://github.com/pulp/pulp-smash/pull/1254

mdellweg: FWIW: RQ, and the cli should not collide on dependencies, because rq is installed inside and the cli outside of the ci-container.
I do however see the cli and smash collide. There could be one way to separate them by installing the cli in a separate venv (maybe using pipx).

fao89: even switching pulp-cli and pulp-smash order doesn't fix the problem,
pulpcore `doc_requirements.txt` brings back click 8.0.0

https://github.com/pulp/pulp_file/runs/2613216940?check_suite_focus=true#step:6:55

mdellweg (approved):

.github/workflows/ci.yml
bmbouter: The pulpcore now does not have a click 8.0.0 requirements. @fao89 I think moving back the changes related to that should now work (as of about 2 hours ago) https://github.com/pulp/pulpcore/pull/1337/files

bmbouter (changes requested):
I wanted to see if the click issue was resolved with no changes in this PR now that https://github.com/pulp/pulpcore/pull/1337/files is merged.

.github/workflows/ci.yml
mdellweg: I'm trying to do away with the click 7 dependency in the cli here:
https://github.com/pulp/pulp-cli/pull/246

fao89:
> I wanted to see if the click issue was resolved with no changes in this PR now that https://github.com/pulp/pulpcore/pull/1337/files is merged.



> I wanted to see if the click issue was resolved with no changes in this PR now that https://github.com/pulp/pulpcore/pull/1337/files is merged.

Look at the commit message, this PR only works because of that PR

bmbouter:
Mmmm I see, ty. Why is the cli block moving? I can also defer to the other reviewers happily, I'm just curious.

fao89:
> Mmmm I see, ty. Why is the cli block moving? I can also defer to the other reviewers happily, I'm just curious.

because I thought pulp-smash needed to be installed first: https://github.com/pulp/plugin_template/pull/389

bmbouter (approved):

PR #508, opened 2021-06-03
Remove FileFilesystemExporter from pulp_file

daviddavis:
fixes #8861


pulp_file/app/viewsets.py
```diff
@@ -184,50 +178,3 @@ class FileDistributionViewSet(DistributionViewSet):
...
     queryset = FileDistribution.objects.all()
     serializer_class = FileDistributionSerializer
-
-
-class FileFilesystemExporterViewSet(ExporterViewSet):
-    """
-    FilesystemExporters export content from a publication to a path on the file system.
-
-    WARNING: This feature is provided as a tech preview and may change in the future. Backwards
-    compatibility is not guaranteed.
```
daviddavis, line 194: :point_up_2:

pulp_file/app/migrations/0012_delete_filefilesystemexporter.py
```diff
@@ -0,0 +1,16 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('file', '0011_fix_auto_publish'),
+    ]
+
+    operations = [
+        migrations.DeleteModel(
+            name='FileFileSystemExporter',
```
mdellweg, line 14: This will only remove the subclassed table, leaving the base table instances around. We might need to add a data migration that removes the objects first.
Django considers this a feature: https://code.djangoproject.com/ticket/32724

daviddavis: I'll add a function to drop them before dropping the table. It looks like the default is to raise a IrreversibleError when reversing it. Is that good or should I just do a noop?

mdellweg: Adding a noop is fine.
OTOH, can we migrate existing exporters to the new home? Than we should move them back in the reverse part.

mdellweg (approved):

bmbouter (approved):
This looks good, thanks!

PR #516, opened 2021-06-14
Moves the milestone check into Python script from bash

dkliban:
The http call was failing due to authentication. The Python script uses the
REDMINE_API_KEY environment variable to authenticate with the API.

[noissue]

### DIFF
--- a/.ci/scripts/redmine.py
+++ b/.ci/scripts/redmine.py
@@ -12,10 +12,15 @@ from redminelib import Redmine
 
 REDMINE_API_KEY = os.environ["REDMINE_API_KEY"]
 REDMINE_QUERY_URL = sys.argv[1]
+MILESTONE_URL = sys.argv[2]
+RELEASE = sys.argv[3]
 CLOSED_CURRENTRELEASE = 11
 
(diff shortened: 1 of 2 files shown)


.ci/scripts/redmine.py
```diff
@@ -12,10 +12,15 @@
 
 REDMINE_API_KEY = os.environ["REDMINE_API_KEY"]
 REDMINE_QUERY_URL = sys.argv[1]
+MILESTONE_URL = sys.argv[2]
+RELEASE = sys.argv[3]
 CLOSED_CURRENTRELEASE = 11
 
 redmine = Redmine(REDMINE_QUERY_URL.split("issues")[0], key=REDMINE_API_KEY)
 query_issues = REDMINE_QUERY_URL.split("=")[-1].split(",")
+milestone_name = redmine.version.get(MILESTONE_URL.split("/")[-1].split(".")[0]).name
```
mdellweg, line 21: Here is an example where we lookup the milestone via the desired release. Not sure, it were applicable here.

https://github.com/pulp/pulp_file/blob/master/.ci/scripts/release.py#L35

PR #522, opened 2021-06-25
Properly check if the tests are being run during release

dkliban:
Since the release is no longer occuring using an existing tag, the old
check does not work.

[noissue]


bmbouter (approved):
Yes we need this. Thanks!

PR #528, opened 2021-07-06
Add a test for syncing from a local repo

dralley:
[noissue]

### DIFF
--- a/pulp_file/tests/functional/api/test_sync.py
+++ b/pulp_file/tests/functional/api/test_sync.py
@@ -1,15 +1,21 @@
-# coding=utf-8
 """Tests that sync file plugin repositories."""
 import unittest
 
 from pulp_smash import config
 from pulp_smash.pulp3.bindings import monitor_task, PulpTaskError
-from pulp_smash.pulp3.utils import gen_repo, get_added_content_summary, get_content_summary
+from pulp_smash.pulp3.utils import (
+    gen_repo,
+    get_added_content_summary,
+    get_content_summary,
+    wget_download_on_host,
+)
 
 from pulp_file.tests.functional.constants import (
-    FILE2_FIXTURE_MANIFEST_URL,
+    FILE_FIXTURE_MANIFEST_URL,
     FILE_FIXTURE_SUMMARY,
+    FILE_FIXTURE_URL,
     FILE_INVALID_MANIFEST_URL,
+    FILE2_FIXTURE_MANIFEST_URL,
 )
 from pulp_file.tests.functional.utils import gen_file_client, gen_file_remote
 from pulp_file.tests.functional.utils import set_up_module as setUpModule  # noqa:F401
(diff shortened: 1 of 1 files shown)


pulp_file/tests/functional/api/test_sync.py
```diff
@@ -32,7 +33,19 @@ def setUpClass(cls):
...
         cls.client = gen_file_client()
 
-    def test_sync(self):
+    @unittest.skipUnless(
+        os.path.exists("/tmp/file_fixture/PULP_MANIFEST"),
+        "Syncing from a file:// repo requires a local fixture to be present.",
+    )
+    def test_sync_local(self):
+        """Test syncing from the local filesystem."""
+        self.do_test("file:///tmp/file_fixture/PULP_MANIFEST")
```
bmbouter, line 42: I had imagined this would come in the pulpcore/tests/using_plugin/ area instead of against pulp_file. Most of the tests there use pulp_file. I don't see the pulpcore CI running pulp_file tests https://github.com/pulp/pulpcore/runs/2995054444?check_suite_focus=true

dralley: I didn't see anywhere in the `using_plugin` tests that directly tested sync, it seemed to fit much better here.

dralley: Tests won't pass until the other PR is merged https://github.com/pulp/pulp_file/pull/528

bmbouter: I hear what you're saying, it's messy at best. The pulpcore using_plugin tests don't directly test sync, but they probably should? Ultimately, that workflow is what validates the entire stages API, downloaders, etc. The guiding principle for me is that while this is an assertion that requires pulp_file to be involved it's actually testing huge amounts of code used by pulpcore.

The concern is that this test put here still won't help us pre-merge. pulpcore will make a breaking change again and merge it. It would help us capture the nightly, but that still wouldn't stop a pulpcore release.

It is messy though, I am not feeling to great about the arrangement here either.

daviddavis: I agree that the using_plugin tests are not optimal. Maybe we can look at running the pulp_file tests against pulpcore? I think we discussed this at one point but decided against it (can't remember why).

I don't have a strong opinion about where this test should live. Since it's to test a fix made in pulpcore, seems like it would be good to have it run against pulpcore.

mdellweg: At least at one point we were running both pulp_file and pulp-certguard tests in the pulpcore ci.

daviddavis: I think @bmbouter said it caused problems when releasing and that's why we removed them from the ci.

dralley: re: discussion last week, my position is that if this happens it should probably happen as part of a greater effort to re-organize tests, because it's definitely not the only test with this issue.

pulp_file/tests/functional/api/test_sync.py
```diff
@@ -32,7 +32,32 @@ def setUpClass(cls):
         cls.cfg = config.get_config()
         cls.client = gen_file_client()
 
-    def test_sync(self):
+    def test_sync_local(self):
+        """Test syncing from the local filesystem."""
+        import shutil
+
+        if not shutil.which("wget"):
+            unittest.skip("Cannot run file:// sync tests without 'wget' available.")
```
mdellweg, line 40: This is not accurate anymore, since we use "wget" on (possibly) another machine.

dralley: By definition it would have to be the same machine, yes?  Or at least with access to the same filesystem?  Because otherwise a `file://` download wouldn't work at all

dralley:
@mdellweg Updated

dralley:
I've opened https://github.com/pulp/pulp-smash/pull/1259 so we can re-use this function across tests / plugins.

mdellweg (approved):

bmbouter:
I'm so glad the pulp-smash utility got added. This test got a lot simpler!

bmbouter:
@dralley The location of the tests between pulp_file and pulpcore are kind of a mess, but for now I'm +1 to merge this.

PR #529, opened 2021-07-08
Drop support for Python 3.6 and 3.7

daviddavis:
This is waiting on pulp/pulp-oci-images#100. I'm waiting for more feedback before I merge the CI Python 3.8 update.

fixes #9037


.github/workflows/scripts/func_test_script.sh
```diff
@@ -19,4 +19,5 @@ then
   fi
 fi
 
+pip install -r ../pulpcore/functest_requirements.txt
```
daviddavis, line 22: Not sure why but it seems like the pulpcore functional tests fail due to being unable to import django. I suspect one of the changes I made to the python dependencies we were installing with dnf when I upgraded the container to python 3.8 had a django dependency.

fao89 (approved):

.github/workflows/scripts/func_test_script.sh
mdellweg: So this line is not from the plugin_template?

daviddavis: No, the func_test_script.sh is [a hook](https://github.com/pulp/pulp_file/blob/5a1b5634ae399c4cbdaf8885e4b2ec38bc10c0ca/.github/workflows/scripts/script.sh#L175-L179) that pulp_file maintains so it can run the pulpcore tests against each PR since changes in pulp_file can break the pulpcore tests.

bmbouter (approved):
This looks great!

PR #535, opened 2021-07-15
Set plugin type on distribution to protect unpublished content

gerrod3:
[noissue]
Required PR: https://github.com/pulp/pulpcore/pull/1495

### DIFF
--- a/pulp_file/app/models.py
+++ b/pulp_file/app/models.py
@@ -124,6 +124,7 @@ class FileDistribution(Distribution):
     """
 
     TYPE = "file"
+    publication_based = True
 
     class Meta:
         default_related_name = "%(app_label)s_%(model_name)s"


ipanova (approved):

dralley (approved):

dralley:
@dkliban This isn't directly a bugfix but it enables a bugfix in pulpcore.

Should @gerrod3 create an issue so that this can be backported to 1.8.z or will there be a 1.9.0 prior to the dropping of support for older Python releases?

I'm slightly in favor of the latter since there is already one other un-backported patch.

gerrod3:
> @dkliban This isn't directly a bugfix but it enables a bugfix in pulpcore.
> 
> Should @gerrod3 create an issue so that this can be backported to 1.8.z or will there be a 1.9.0 prior to the dropping of support for older Python releases?
> 
> I'm slightly in favor of the latter since there is already one other un-backported patch.

Backporting this would only make sense if the fix in pulpcore is also backported to 3.13 and 3.14. Pulp_file 1.8 has pulpcore requirements >=3.13,<3.15.

dralley:
It probably should be backported to 3.14.

PR #536, opened 2021-07-16
Acs refresh

pavelpicka:
Added `refresh` endpoint to support ACS.

closes: #9377
https://pulp.plan.io/issues/9377

Required PR: https://github.com/pulp/pulpcore/pull/1496

### DIFF
--- a/docs/workflows/alternate-content-source.rst
+++ b/docs/workflows/alternate-content-source.rst
@@ -1,0 +1,29 @@
+Alternate Content Sources
+=========================
+
+To use an Alternate Content Source you need a ``FileRemote`` with base path of your
+ACS.
+
+.. code-block:: bash
+
+    http POST localhost:24817/pulp/api/v3/remotes/file/file/ name="remoteForACS" policy="on_demand" url="http://fixtures.pulpproject.org/"
+
+Create Alternate Content Source
+-------------------------------
+
+.. code-block:: bash
+
+    http POST localhost:24817/pulp/api/v3/acs/file/file/ name="fileAcs" remote=<pulp-remote-href> paths:='["file/PULP_MANIFEST", "backup/MANIFEST"]'
+
+Refresh ACS
+-----------
+
+To make your ACS available for future syncs you need to call ``refresh`` endpoint
+on your ACS.
+
+.. code-block:: bash
+
+    http POST localhost:24817/pulp/api/v3/acs/file/file/<ACS-UUID>/refresh/
+
+Alternate Content Source has a global scope so if any content is found in ACS it
+will be used in all future syncs.
--- a/pulp_file/app/tasks/synchronizing.py
+++ b/pulp_file/app/tasks/synchronizing.py
@@ -24,7 +24,7 @@ log = logging.getLogger(__name__)
 metadata_files = []
 
 
-def synchronize(remote_pk, repository_pk, mirror):
+def synchronize(remote_pk, repository_pk, mirror, url=None):
     """
     Sync content from the remote repository.
 
@@ -34,6 +34,7 @@ def synchronize(remote_pk, repository_pk, mirror):
         remote_pk (str): The remote PK.
         repository_pk (str): The repository PK.
         mirror (bool): True for mirror mode, False for additive.
+        url (str): The url to synchronize. If omitted, the url of the remote is used.
 
     Raises:
         ValueError: If the remote does not specify a URL to sync.
@@ -45,8 +46,8 @@ def synchronize(remote_pk, repository_pk, mirror):
     if not remote.url:
         raise ValueError(_("A remote must have a url specified to synchronize."))
 
-    first_stage = FileFirstStage(remote)
-    dv = DeclarativeVersion(first_stage, repository, mirror=mirror)
+    first_stage = FileFirstStage(remote, url)
+    dv = DeclarativeVersion(first_stage, repository, mirror=mirror, acs=True)
     rv = dv.create()
     if rv and mirror:
         # TODO: this is awful, we really should rewrite the DeclarativeVersion API to
@@ -72,16 +73,18 @@ class FileFirstStage(Stage):
     The first stage of a pulp_file sync pipeline.
     """
 
-    def __init__(self, remote):
+    def __init__(self, remote, url):
         """
         The first stage of a pulp_file sync pipeline.
 
         Args:
             remote (FileRemote): The remote data to be used when syncing
+            url (str): The base url of custom remote
 
         """
         super().__init__()
         self.remote = remote
+        self.url = url if url else remote.url
 
     async def run(self):
         """
@@ -93,12 +96,12 @@ class FileFirstStage(Stage):
         async with ProgressReport(
             message="Downloading Metadata", code="sync.downloading.metadata"
         ) as pb:
-            parsed_url = urlparse(self.remote.url)
+            parsed_url = urlparse(self.url)
             root_dir = os.path.dirname(parsed_url.path)
-            downloader = self.remote.get_downloader(url=self.remote.url)
+            downloader = self.remote.get_downloader(url=self.url)
             result = await downloader.run()
             await pb.aincrement()
-            metadata_files.append((result.path, self.remote.url.split("/")[-1]))
+            metadata_files.append((result.path, self.url.split("/")[-1]))
 
         async with ProgressReport(
             message="Parsing Metadata Lines", code="sync.parsing.metadata"
--- a/pulp_file/app/viewsets.py
+++ b/pulp_file/app/viewsets.py
@@ -1,11 +1,19 @@
+import os
+
 from django_filters import CharFilter
 from drf_spectacular.utils import extend_schema
 from rest_framework.decorators import action
 
 from pulpcore.plugin.actions import ModifyRepositoryActionMixin
+from pulpcore.plugin.models import (
+    AlternateContentSource,
+    AlternateContentSourcePath,
+    TaskGroup,
+)
 from pulpcore.plugin.serializers import (
     AsyncOperationResponseSerializer,
     RepositorySyncURLSerializer,
+    TaskGroupResponseSerializer,
 )
 from pulpcore.plugin.tasking import dispatch
 from pulpcore.plugin.viewsets import (
@@ -18,6 +26,7 @@ from pulpcore.plugin.viewsets import (
     RepositoryViewSet,
     RepositoryVersionViewSet,
     SingleArtifactContentUploadViewSet,
+    TaskGroupResponse,
 )
 
 from . import tasks
@@ -194,3 +203,50 @@ class FileAlternateContentSourceViewSet(AlternateContentSourceViewSet):
     endpoint_name = "file"
     queryset = FileAlternateContentSource.objects.all()
     serializer_class = FileAlternateContentSourceSerializer
+
+    @extend_schema(
+        description="Trigger an asynchronous task to create Alternate Content Source content.",
+        responses={202: TaskGroupResponseSerializer},
+    )
+    @action(methods=["post"], detail=True)
+    def refresh(self, request, pk):
+        """
+        Refresh ACS metadata.
+        """
+        acs = AlternateContentSource.objects.get(pk=pk)
+        acs_paths = AlternateContentSourcePath.objects.filter(alternate_content_source=pk)
+        task_group = TaskGroup.objects.create(
+            description=f"Refreshing {acs_paths.count()} alternate content source paths."
+        )
+
+        for acs_path in acs_paths:
+            # Create or get repository for the path
+            repo_data = {
+                "name": f"{acs.name}--{acs_path.pk}--repository",
+                "retain_repo_versions": 1,
+                "user_hidden": True,
+            }
+            repo, created = FileRepository.objects.get_or_create(**repo_data)
+            if created:
+                acs_path.repository = repo
+                acs_path.save()
+            acs_url = (
+                os.path.join(acs.remote.url, acs_path.path) if acs_path.path else acs.remote.url
+            )
+
+            # Dispatching ACS path to own task and assign it to common TaskGroup
+            dispatch(
+                tasks.synchronize,
+                shared_resources=[acs.remote, acs],
+                task_group=task_group,
+                kwargs={
+                    "remote_pk": str(acs.remote.pk),
+                    "repository_pk": str(acs_path.repository.pk),
+                    "mirror": False,
+                    "url": acs_url,
+                },
+            )
+
+        # Update TaskGroup that all child task are dispatched
+        task_group.finish()
+        return TaskGroupResponse(task_group, request)
--- a/CHANGES/9377.feature
+++ b/CHANGES/9377.feature
@@ -1,0 +1,1 @@
+Added refresh endpoint for Alternate Content Sources.
--- a/docs/workflows/index.rst
+++ b/docs/workflows/index.rst
@@ -41,3 +41,4 @@ library with:
    sync
    upload
    publish-host
+   alternate-content-source


CHANGES/9097.feature
```diff
@@ -0,0 +1 @@
+Added support for Alernate Content Sources.
```
daviddavis, line 1: I thought there was already support for ACS? Maybe this should say added a refresh endpoint?

docs/workflows/alternate-content-source.rst
```diff
@@ -0,0 +1,33 @@
+Alternate Content Sources
+=========================
+
+To use an Alternate Content Source you need a ``remote`` with base path of your
+ACS ending with ``MANIFEST`` filename.
+
+.. code-block:: bash
+
+    http --form POST localhost:24817/pulp/api/v3/remotes/file/file/ name="fremote" policy="on_demand" url="http://fixtures.pulpproject.org/PULP_MANIFEST"
```
daviddavis, line 9:
```suggestion
    http POST localhost:24817/pulp/api/v3/remotes/file/file/ name="remote" policy="on_demand" url="http://fixtures.pulpproject.org/PULP_MANIFEST"
```

docs/workflows/alternate-content-source.rst
```diff
@@ -0,0 +1,33 @@
...
+
+This manifest filename will be used for all paths under same ACS.  E.g. your
+path is ``file/`` final url where to look for alternate content will be
+``http://fixtures.pulpproject.org/file/PULP_MANIFEST``.
```
daviddavis, line 13: This seems strange but I suppose it's a byproduct of having remote urls contain the name of the manifest?

docs/workflows/alternate-content-source.rst
```diff
@@ -0,0 +1,33 @@
...
+
+To use an Alternate Content Source you need a ``remote`` with base path of your
+ACS ending with ``MANIFEST`` filename.
+
+.. code-block:: bash
+
+    http --form POST localhost:24817/pulp/api/v3/remotes/file/file/ name="fremote" policy="on_demand" url="http://fixtures.pulpproject.org/PULP_MANIFEST"
+
+This manifest filename will be used for all paths under same ACS.  E.g. your
+path is ``file/`` final url where to look for alternate content will be
```
daviddavis, line 12:
```suggestion
This manifest filename will be used for all paths under same ACS.  For example,
if a path is ``file/`` the final url where Pulp will look for alternate content will be
```

pulp_file/app/tasks/acs.py
```diff
@@ -0,0 +1,45 @@
...
+def file_acs_refresh(acs_pk):
+    """
+    Refresh alternative content sources.
+    """
+    acs = AlternateContentSource.objects.get(pk=acs_pk)
+    paths_query = len(AlternateContentSourcePath.objects.filter(alternate_content_source=acs_pk))
+    with ProgressReport(
+        message="Refreshing ACS metadata", code="acs.refreshing.metadata", total=paths_query
+    ) as pb:
+        for acs_path in AlternateContentSourcePath.objects.filter(alternate_content_source=acs_pk):
```
daviddavis, line 27:
```suggestion
    acs_paths = AlternateContentSourcePath.objects.filter(alternate_content_source=acs_pk)
    with ProgressReport(
        message="Refreshing ACS metadata", code="acs.refreshing.metadata", total=len(acs_paths)
    ) as pb:
        for acs_path in acs_paths:
```

docs/workflows/alternate-content-source.rst
pavelpicka: This is a specific of file plugin. It required the url with manifest in the url.

daviddavis: Maybe we should have the remote url point to `http://fixtures.pulpproject.org/` and the paths point to `file/PULP_MANIFEST`? This seems less strange than having the remote point to `http://fixtures.pulpproject.org/PULP_MANIFEST` and the path point to `file/` and would also allow for different manifest file names per path.

bmbouter: > Maybe we should have the remote url point to `http://fixtures.pulpproject.org/` and the paths point to `file/PULP_MANIFEST`? This seems less strange than having the remote point to `http://fixtures.pulpproject.org/PULP_MANIFEST` and the path point to `file/` and would also allow for different manifest file names per path.

I agree

daviddavis: If we make this change, we'll have to remove the validation that checks that paths end in `/`. This happens in pulpcore but I think it makes sense to move validation of paths to the plugins:

https://github.com/pulp/pulpcore/pull/1607
https://github.com/pulp/pulp_file/pull/569

pulp_file/app/viewsets.py
```diff
@@ -193,3 +193,18 @@ class FileAlternateContentSourceViewSet(AlternateContentSourceViewSet):
     endpoint_name = "file"
     queryset = FileAlternateContentSource.objects.all()
     serializer_class = FileAlternateContentSourceSerializer
+
+    @action(methods=["post"], detail=True)
+    def refresh(self, request, pk):
```
daviddavis, line 198: I think you'll need to add `@extend_schema` here as the schema currently indicates that a 201 will be returned but a 202 actually gets returned.

pulp_file/app/tasks/acs.py
```diff
@@ -0,0 +1,46 @@
...
+                "name": f"{acs.name}--{acs_path.pk}--repository",
+                "retain_repo_versions": 1,
+                "user_hidden": True,
+            }
+            repo, created = FileRepository.objects.get_or_create(**repo_data)
+            if created:
+                acs_path.repository = repo
+                acs_path.save()
+            acs_url = os.path.join(acs.remote.url, acs_path.path)
+            synchronize(acs.remote.pk, repo.pk, False, acs_url)
```
bmbouter, line 44: Instead of calling `synchronize` directly here dispatch this task instead:  https://github.com/pulp/pulp_file/blob/main/pulp_file/app/tasks/synchronizing.py#L27  You can do that similarly to how the sync viewset does:  https://github.com/pulp/pulp_file/blob/main/pulp_file/app/viewsets.py#L98-L107

The main difference is that you want all of your tasks to be a task group. You should create a `TaskGroup` and then set it on each task dispatched https://github.com/pulp/pulpcore/blob/24be668caed280eeda7413a563cdd612d423441c/pulpcore/tasking/tasks.py#L45

Here's an example of TaskGroup being created https://github.com/pulp/pulpcore/blob/de67d41c39d9d4bde2e8f86e4bf3f51d6cc774e7/pulpcore/app/tasks/importer.py#L376 Also list it as a CreatedResource https://github.com/pulp/pulpcore/blob/de67d41c39d9d4bde2e8f86e4bf3f51d6cc774e7/pulpcore/app/tasks/importer.py#L379  Then set the TaskGroup on the tasks dispatched.

I'd recommend not doing any `GroupProgressReport` for now, and instead we can add that in after all ^ is done.

Post any questions you have please. Thanks!

docs/workflows/alternate-content-source.rst
```diff
@@ -0,0 +1,29 @@
+Alternate Content Sources
+=========================
+
+To use an Alternate Content Source you need a ``remote`` with base path of your
```
bmbouter, line 4:
```suggestion
To use an Alternate Content Source you need a ``FileRemote`` with base path of your
```

docs/workflows/alternate-content-source.rst
```diff
@@ -0,0 +1,29 @@
...
+ACS.
+
+.. code-block:: bash
+
+    http POST localhost:24817/pulp/api/v3/remotes/file/file/ name="remoteForACS" policy="on_demand" url="http://fixtures.pulpproject.org/"
```
bmbouter, line 9: I think we should use pulp CLI commands instead.

docs/workflows/alternate-content-source.rst
```diff
@@ -0,0 +1,29 @@
...
+
+Create Alternate Content Source
+-------------------------------
+
+.. code-block:: bash
+
+    http POST localhost:24817/pulp/api/v3/acs/file/file/ name="fileAcs" remote=<pulp-remote-href> paths:='["file/PULP_MANIFEST", "backup/MANIFEST"]'
```
bmbouter, line 16: This one can stay as httpie until we have CLI support for it and then switch it. (I think)

pulp_file/app/tasks/acs.py
```diff
@@ -0,0 +1,46 @@
...
+from pulp_file.app.models import FileRepository
+
+metadata_files = []
+
+
+@extend_schema(
+    description="Trigger an asynchronous task to create Alternate Content Source content.",
+    responses={202: AsyncOperationResponseSerializer},
+)
+def file_acs_refresh(acs_pk):
```
bmbouter, line 23: I think most of this should be in a viewset and dispatch to the N subtasks group. For example the `extend_schema` stuff is all viewset stuff.

pulp_file/app/tasks/synchronizing.py
```diff
@@ -72,16 +73,18 @@ class FileFirstStage(Stage):
...
         The first stage of a pulp_file sync pipeline.
 
         Args:
             remote (FileRemote): The remote data to be used when syncing
+            url (str): The base url of custom remote
 
         """
         super().__init__()
         self.remote = remote
+        self.url = url if url else remote.url
```
bmbouter, line 87: :+1:

pulp_file/app/tasks/synchronizing.py
```diff
@@ -34,6 +34,7 @@ def synchronize(remote_pk, repository_pk, mirror):
         remote_pk (str): The remote PK.
         repository_pk (str): The repository PK.
         mirror (bool): True for mirror mode, False for additive.
+        url (str): The custom url of remote
```
bmbouter, line 37:
```suggestion
        url (str): The url to synchronize. If omitted, the url of the remote is used.
```

pulp_file/app/tasks/synchronizing.py
bmbouter: Maybe moving this to `__init__` would make it more explicit (because it happens earlier)

docs/workflows/alternate-content-source.rst
daviddavis: The CLI commands haven’t been merged yet so they’re subject to changing. I’d be happy to update this in a follow up PR.

pulp_file/app/tasks/synchronizing.py
pavelpicka: Which ```__init__``` do you mean? I believe this is first place where it is needed to be checked.

pulp_file/app/tasks/acs.py
pavelpicka: I agree that ```extend_schema``` must go to the viewset, but other logic I'd keep out of viewset as there are happening many db queries. So to keep viewset fast.

bmbouter: That sounds ok. It will mean you'll have a task that dispatches other tasks and that is ok too. Is it clear to you what you should do to make this?

pulp_file/app/tasks/synchronizing.py
bmbouter: My comment didn't make sense. I was meant to ask if it made sense to move this to `synchronize()` because then the value of `url` would be more concrete earlier in the call stack. Just a thought. Do what you think makes sense.

pulp_file/app/tasks/acs.py
bmbouter: I reread this diff and it seems you've already done it. :+1:

pulp_file/app/tasks/acs.py
```diff
@@ -0,0 +1,55 @@
...
+    """
+    acs = AlternateContentSource.objects.get(pk=acs_pk)
+    acs_paths = AlternateContentSourcePath.objects.filter(alternate_content_source=acs_pk)
+    with ProgressReport(
+        message="Refreshing ACS metadata", code="acs.refreshing.metadata", total=len(acs_paths)
+    ) as pb:
+        for acs_path in acs_paths:
+            # Create or get repository for each path
+            repo_data = {
+                "name": f"{acs.name}--{acs_path.pk}--repository",
```
bmbouter, line 31: Is this logic referenced elsewhere anywhere? Or is this the one place this string is formed?

pavelpicka: This is only place. Intention is to have unique name.

pulp_file/app/tasks/acs.py
pavelpicka: Great, I was confused that some part logic should be moved as well.

pulp_file/app/viewsets.py
```diff
@@ -194,3 +198,18 @@ class FileAlternateContentSourceViewSet(AlternateContentSourceViewSet):
...
+        """
+        Refresh ACS metadata.
+        """
+        acs = self.get_object()
+        result = dispatch(
+            tasks.file_acs_refresh,
+            [acs],
```
ipanova, line 210: maybe use exclusive locks

pulp_file/app/tasks/acs.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+def file_acs_refresh(acs_pk):
+    """
+    Refresh alternative content sources.
+    """
+    acs = AlternateContentSource.objects.get(pk=acs_pk)
+    acs_paths = AlternateContentSourcePath.objects.filter(alternate_content_source=acs_pk)
+    with ProgressReport(
+        message="Refreshing ACS metadata", code="acs.refreshing.metadata", total=len(acs_paths)
```
bmbouter, line 26: This should likely identify that it's refreshing N repos.
```suggestion
        message="Dispatching tasks to refresh metadata", code="acs.refresh.dispatch", total=len(acs_paths)
```

pulp_file/app/tasks/acs.py
```diff
@@ -0,0 +1,55 @@
...
+                "retain_repo_versions": 1,
+                "user_hidden": True,
+            }
+            repo, created = FileRepository.objects.get_or_create(**repo_data)
+            if created:
+                acs_path.repository = repo
+                acs_path.save()
+            acs_url = os.path.join(acs.remote.url, acs_path.path)
+            # Dispatch each ACS path to own task and assign it to common TaskGroup
+            task_group = TaskGroup.objects.create(description=f"Refresh of {acs_path.path}")
```
bmbouter, line 41: We won't be able to move the TaskGroup creation to the viewset until [we implement this in pulpcore](https://pulp.plan.io/issues/9380). Let's continue to keep the `TaskGroup` creation in this task for now until we can do that and then we'll refactor it. I plan to discuss this change at the pulpcore meeting next Tuesday.

pulp_file/app/tasks/acs.py
```diff
@@ -0,0 +1,55 @@
...
+                exclusive_resources=[acs],
+                task_group=task_group,
+                kwargs={
+                    "remote_pk": str(acs.remote.pk),
+                    "repository_pk": str(acs_path.repository.pk),
+                    "mirror": False,
+                    "url": acs_url,
+                },
+            )
+            pb.increment()
```
bmbouter, line 54: You need to add a CreatedResource call to have this task identify the TaskGroup as something that was created. See how import [does it here](https://github.com/pulp/pulpcore/blob/de67d41c39d9d4bde2e8f86e4bf3f51d6cc774e7/pulpcore/app/tasks/importer.py#L379) as an example.

pulp_file/app/tasks/acs.py
bmbouter: I made this https://github.com/pulp/pulpcore/pull/1618/files

bmbouter: @pavelpicka this is now available for use in pulpcore https://github.com/pulp/pulpcore/pull/1618

So I think move the TaskGroup creation to the viewset and return it to the user using the new Response type in pulpcore. Then have all tasks (including this one and its subtasks) be members of the task_group.

pulp_file/app/viewsets.py
```diff
@@ -194,3 +207,45 @@ class FileAlternateContentSourceViewSet(AlternateContentSourceViewSet):
...
+                task_group=task_group,
+                kwargs={
+                    "remote_pk": str(acs.remote.pk),
+                    "repository_pk": str(acs_path.repository.pk),
+                    "mirror": False,
+                    "url": acs_url,
+                },
+            )
+            CreatedResource(content_object=task)  # No effect on http responses
+        CreatedResource(content_object=task_group)  # No effect on http responses
```
bmbouter, line 249: Yeah you won't need these anymore.

bmbouter: After all tasks have been dispatched, you should issue set `all_tasks_dispatched` to `True` https://github.com/pulp/pulpcore/blob/24be668caed280eeda7413a563cdd612d423441c/pulpcore/app/models/task.py#L283

This solves the problem where if somehow a client did see the TaskGroup and all the current tasks were completed, it could still know that more were going to be created. Otherwise it couldn't tell the difference between "I've done everything so far" versus "I've done everything".

pulp_file/app/viewsets.py
```diff
@@ -194,3 +207,45 @@ class FileAlternateContentSourceViewSet(AlternateContentSourceViewSet):
...
+
+        return TaskGroupResponse(task_group, request)
```
bmbouter, line 251: :+1:

pulp_file/app/viewsets.py
```diff
@@ -184,6 +193,10 @@ class FileDistributionViewSet(DistributionViewSet):
     serializer_class = FileDistributionSerializer
 
 
+@extend_schema(
+    description="Trigger an asynchronous task to create Alternate Content Source content.",
+    responses={202: AsyncOperationResponseSerializer},
```
bmbouter, line 198: This needs to use the `TaskGroupResponseSerializer`.

pulp_file/app/viewsets.py
```diff
@@ -194,3 +207,46 @@ class FileAlternateContentSourceViewSet(AlternateContentSourceViewSet):
...
+            if created:
+                acs_path.repository = repo
+                acs_path.save()
+            acs_url = os.path.join(acs.remote.url, acs_path.path)
+
+            # Dispatching ACS path to own task and assign it to common TaskGroup
+            dispatch(
+                tasks.synchronize,
+                shared_resources=[acs.remote],
+                exclusive_resources=[acs],
```
bmbouter, line 239: I believe this will cause all N sync tasks to be run serially as each one waits for the exclusive lock on `acs`. I think this should be a `shared_resource` too?

pavelpicka: It should!

docs/workflows/alternate-content-source.rst
```diff
@@ -0,0 +1,29 @@
...
+-----------
+
+To make your ACS available for future syncs you need to call ``refresh`` endpoint
+on your ACS.
+
+.. code-block:: bash
+
+    http POST localhost:24817/pulp/api/v3/acs/file/file/<ACS-UUID>/refresh/
+
+Alternate Content Source has a global scope so if any content found in ACS it
```
ipanova, line 28: s/content found/content is found/

pulp_file/app/viewsets.py
```diff
@@ -194,3 +203,49 @@ class FileAlternateContentSourceViewSet(AlternateContentSourceViewSet):
...
+                    "remote_pk": str(acs.remote.pk),
+                    "repository_pk": str(acs_path.repository.pk),
+                    "mirror": False,
+                    "url": acs_url,
+                },
+            )
+
+        # Update TaskGroup that all child task are dispatched
+        task_group.all_tasks_dispatched = True
+        task_group.save()
```
ipanova, line 250: i'd call task_group.finish() instead.

ipanova:
The code changes look good to me :+1:

pulp_file/app/viewsets.py
```diff
@@ -194,3 +203,49 @@ class FileAlternateContentSourceViewSet(AlternateContentSourceViewSet):
...
+            repo_data = {
+                "name": f"{acs.name}--{acs_path.pk}--repository",
+                "retain_repo_versions": 1,
+                "user_hidden": True,
+            }
+            repo, created = FileRepository.objects.get_or_create(**repo_data)
+            if created:
+                acs_path.repository = repo
+                acs_path.save()
+            acs_url = os.path.join(acs.remote.url, acs_path.path)
```
daviddavis, line 233: I think this should check if path is blank or not. I'm imagining that if you don't set paths for the ACS, it'll just use the remote path (e.g. `http://fixtures.pulpproject.org/PULP_MANIFEST`) but what it does instead is it appends a slash (`http://fixtures.pulpproject.org/PULP_MANIFEST/`).

pavelpicka: but not here as this is in the loop through paths, it is needed to add a logic when no paths at all

pavelpicka: sounds like good idea to add tests for no-path ACS and ACS with more paths.

daviddavis: Good idea, will do 👍

daviddavis (approved):

PR #538, opened 2021-07-16
Fixed 0012_delete_filefilesystemexporter migration

daviddavis:
fixes #9102

### DIFF
--- a/CHANGES/9102.bugfix
+++ b/CHANGES/9102.bugfix
@@ -1,0 +1,2 @@
+Fixed failing 0012_delete_filefilesystemexporter migration which errors if there are
+``FileFilesystemExporters``.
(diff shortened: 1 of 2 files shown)


daviddavis:
I only know of this fix because of the distribution migration (see below) and I am not totally sure why it works.

https://github.com/pulp/pulp_file/blob/d0dd4ebd6f2c960db3441d572ba6f825210f0e03/pulp_file/app/migrations/0009_move_data_to_new_master_distribution_model.py#L48

dralley:
Adding @mdellweg as a reviewer since he wrote the original patch

mdellweg:
I only have a theory:
In this other migration, a table got renamed. And at the end of the transaction, some triggers were pending on the old table (-name) and failed. Setting `atomic=False` end the transactions earlier, and thereby triggers triggers earlier.

mdellweg (approved):
Here the deleting of objects will probably add some triggers on the table that is removed in the next step.
The only other solution might be to split the migration in two.

PR #544, opened 2021-07-27
Touch content to prevent cleanup

daviddavis:
fixes #9154

### DIFF
--- a/pulp_file/app/serializers.py
+++ b/pulp_file/app/serializers.py
@@ -38,6 +38,8 @@ class FileContentSerializer(SingleArtifactContentUploadSerializer, ContentChecks
         )
 
         if content.exists():
+            content.get().touch()
+
             raise serializers.ValidationError(
                 _(
                     "There is already a file content with relative path '{path}' and digest "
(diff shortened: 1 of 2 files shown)


daviddavis:
There was one other place that seems like a `touch()` call might make sense:

https://github.com/pulp/pulp_file/blob/c6300d1e0460ed1273ac31114cccbc4c49683eb7/pulp_file/app/tasks/publishing.py#L67-L69

However, this content must belong to a repo version and therefore it wouldn't be orphaned.

ipanova (approved):

pulp_file/app/serializers.py
```diff
@@ -38,6 +38,9 @@ def deferred_validate(self, data):
         )
 
         if content.exists():
+            for content in content.iterator():
```
ipanova, line 41: +1

mdellweg (approved):

pulp_file/app/serializers.py
mdellweg: We do not expect more than one item here, right? ;)

ipanova: we don't, but i think iterator will not bring any harm..

daviddavis: I like `content.get().touch()`. Will update.

PR #546, opened 2021-07-28
Enable reclaim disk feature provided by pulpcore.

ipanova:
closes #9168
https://pulp.plan.io/issues/9168

Required PR: https://github.com/pulp/pulpcore/pull/1452

### DIFF
--- a/CHANGES/9168.feature
+++ b/CHANGES/9168.feature
@@ -1,0 +1,1 @@
+Enable reclaim disk feature provided by pulpcore. This feature is available with pulpcore 3.15+.
--- a/pulp_file/app/models.py
+++ b/pulp_file/app/models.py
@@ -28,6 +28,8 @@ class FileContent(Content):
         digest (str): The SHA256 HEX digest.
     """
 
+    PROTECTED_FROM_RECLAIM = False
+
     TYPE = "file"
     repo_key_fields = ("relative_path",)
 


pulp_file/app/tasks/publishing.py
```diff
@@ -74,9 +74,10 @@ def populate(publication):
         else:
             # TODO: this scales poorly, one query per on_demand content being published.
             artifact = RemoteArtifact.objects.filter(content_artifact=content_artifact).first()
-        entry = Entry(
-            relative_path=content_artifact.relative_path,
-            digest=artifact.sha256,
-            size=artifact.size,
-        )
-        yield entry
+            if artifact:
```
ipanova, line 77: should we father fail publish or producing empty pulp-manifest is ok?

pulp_file/app/tasks/publishing.py
```diff
@@ -74,9 +74,10 @@ def populate(publication):
...
-            size=artifact.size,
-        )
-        yield entry
+        if artifact:
+            entry = Entry(
+                relative_path=content_artifact.relative_path,
+                digest=artifact.sha256,
+                size=artifact.size,
+            )
+            yield entry
```
mdellweg, line 83: I didn't see this change coming.
If there is no such artifact, does that mean, the content is broken? I think we should deny creating a publication from it. Or, at the very least put out a big warning, that something is missing and you'd want to repair your content.
Silently publishing whatever is left in the repository version doesn't sound right to me.

ipanova: This was my question in the earlier comment. I means that content does not have artifact nor remote artifact because it was uploaded and later on reclaimed. Such content is un-repairable. As a result, if there is a further desire to publish such content there will be an empty pulp_manifest.

mdellweg: I believe that we should bare with the fact that unrepairable content is broken, and that repository versions containing broken content are broken too. And even publications made before forceful reclaiming are in a way broken. Their metadata will point to content that is no longer downloadable.
One other way, i think would be ok, were to create a pulp_manifest including the information about missing artifacts. I don't know if we have that information though.

ipanova: The only information available to us would the rel_path which we can include in the pulp-manifest.
To be clear, these changes are made not to make the publication usable for not-usable content but to be able to handle that task and not fail with 500

mdellweg: i think digest is also on the content table. But still size is missing.
I think producing the pulp_manifest with silently skipping broken content is lying to the user.

dralley: >And even publications made before forceful reclaiming are in a way broken. Their metadata will point to content that is no longer downloadable.

This is somewhat expected, it's a risk you take any time you use `on_demand`.  As long as we're returning 404s instead of 500s it's probably OK.

As a short-term solution I'm not against @ipanova's implementation, but over the medium term I mostly agree with @mdellweg, I just don't think we have time to implement it now.  If we implemented a hard error, we will still need to think about the user experience, for example we probably don't want to make the user manually publish over and over, each time manually removing impacted content units from their repository.  Even though the user opted to use `--force` and probably "deserves it" (lol), that would be a very painful situation to recover from.

Something like [this](https://pulp.plan.io/issues/8305#note-15) would help, but only after a new repository version was created. But the scope is now rapidly growing beyond this one story.

What are your thoughts on this @mdellweg?

This potentially has some overlap with the problem of known-bad or malicious packages that should be "purged" from Pulp permanently.  So if we want to consider comprehensive solutions, there are actual use cases that might justify and benefit from it, and I do think we should address this properly.

mdellweg: I know that we give the user the `--force` to create defunct content. OTOH the only thing missing on the file content table to create valid metadata even for missing artifacts is the size, which atm we need to derive from the `{Remote,}Artifact`. We could easily add that column to the file content and write a content migration for it. And with that we would be able to create the same publication of a repository version before and after using the force reclaim. And that would be in the same (miserable) state as a publication with on-demand content and removed remote. I think it would be desirable to move these two situation into the same corner case.
In general it would be stated as: A plugin needs to put all information to create publication metadata into the content model. And i think this is a valid assumption, if we drop the assumption that content can always assume to have a certain set of `ContentArtifact`s, which is an unavoidable side effect of the whole reclaim story.

BTW: This might have a positive impact on publication time.

goosemania: I do not think it's helpful to publish broken content. We know it can't be used.
I suggest to accept the current proposal and add the `else` clause with the warning and not add anything to a pulp_manifest.

ipanova: I want to point out that pulp-rpm publishes now everything and without any sort of warning

goosemania: If so, I suggest to fix it there. cc @dralley

dralley: `--force` is such a headache.  @ipanova how difficult would it be to "fix" there in this same way?

ipanova: @dralley @goosemania @mdellweg I'd like to finalize this PR so pulpcore ci passes since its tests depend on the file plugin. When a publication is created in pulp_file plugin, the "broken" content is not included, it's omitted, so for example, if the repo version contains 10 units and only 1 is ok ( has artifact or remote artifact) manifest file will have just one record. If none of the content is ok, the manifest will be empty.
At this point I can whether a) do nothing b) create a warning with generic message which will say that during publish some broken content was identified and thus omitted  c) create a warning which will contain a specific message which will provide a list of content (rel_path or digest?) for the content that was omitted from publish.
Preferences?
@dralley for for rpm plugin question i will look into tomorrow.

mdellweg: The only information that is missing on the `FileContent` model to create a "valid publication with missing content" is the `size` attribute. And that _was_ ok in the Pulp model, because Pulp promised that a `FileContent` object was always associated with a `{Remote,}Artifact`. That promise is no longer true, once we switch on the reclaim capability. My point is, now is the very last chance to write a migration that copies the `size` attribute into the `FileContent` before anyone can create broken content we can never repair anymore. And I want us to do just that.

Somehow, for the sole reason of being in the natural key, the sha256 was always on the `FileContent` model in form of the `digest` attribute.

ipanova:
the `force` flag has been withdrawn from the api. once this pr in core is merged i believe pulp-file can merge this one and it will be ready for the release https://github.com/pulp/pulpcore/pull/1544
@dkliban fyi

dralley (approved):

pulp_file/app/tasks/publishing.py
goosemania: @mdellweg, could you share what's the point of publishing broken content?
If you publish it and distribute, you'll make it "available"/visible for clients to try download. It would lead only to frustrations and pointless debugging of why it's not working, would you agree?

If it's not published/distributed and a warning is logged, a user is aware and can either ignore it or make sure this content is fixed.
When and if the content is fixed, it will appear in the subsequent publication.
My only concern with the --force approach is, can a user fix the broken content by uploading? Or will pulp complain that such content already exists and do nothing?

ipanova: @goosemania I had a short meeting @dralley yesterday to outline my observations but I'd like to share them here as well.
Our team has basically divide into 2 camps with different philosophy. One camp believes that we need to keep the promise of 'Whatever is present in the repo version will be 100% distributed, but client might get 404' vs 'Pulp does extra work and checks to ensure only sane content comes down to the client requests'.
Both approaches would raise questions like"I have 10 rpms in my repo but i get 404 for 8 of then, how so?" or "I have 10 rpms in my repo but only 2 were published, where is the rest?"
That's what prompted me to withdraw the force options since I don't think we can find a common ground of this question only by discussing it on the PR.

ipanova: And that's only for plugins that have publications, the ones that have  distributions only will not have the 'opportunity' to select and distribute only sane content so those plugins would always distribute with 100% match what the repo version has.

goosemania: > And that's only for plugins that have publications, the ones that have distributions only will not have the 'opportunity' to select and distribute only sane content so those plugins would always distribute with 100% match what the repo version has.

Oh that's a valid point and a pity then.
I guess for consistency, we should publish everything :(

goosemania: > @goosemania I had a short meeting @dralley yesterday to outline my observations but I'd like to share them here as well.

Thank you!

> Our team has basically divide into 2 camps with different philosophy. One camp believes that we need to keep the promise of 'Whatever is present in the repo version will be 100% distributed, but client might get 404' vs 'Pulp does extra work and checks to ensure only sane content comes down to the client requests'.
> Both approaches would raise questions like"I have 10 rpms in my repo but i get 404 for 8 of then, how so?" or "I have 10 rpms in my repo but only 2 were published, where is the rest?"

That one is easy for me. The rest you see in the warnings in the logs.

> That's what prompted me to withdraw the force options since I don't think we can find a common ground of this question only by discussing it on the PR.

Thanks

mdellweg: > > And that's only for plugins that have publications, the ones that have distributions only will not have the 'opportunity' to select and distribute only sane content so those plugins would always distribute with 100% match what the repo version has.
> 
> Oh that's a valid point and a pity then.
> I guess for consistency, we should publish everything :(

To me, this is even true for publications of deferred artifacts. If upstream decides to stop serving a file, the only thing we can do is return "404 Sorry not there anymore".

OTOH consider a publication of a repoversion created with dependency solving. If the particular artifacts are missing and we stop publishing them, dependencies aren't resolved anymore, and the metadata could be considered invalid in that respect.

If however we can repair such content by reuploading artifacts, we can "repair" publications that are broken in that respect.

> @mdellweg, could you share what's the point of publishing broken content?

My thinking is that a repository versions is immutable, so it should not matter when i create a publication from it. Be it prior to a reclaim, or afterwards. And in my experience, packaging tools like yum, pip, apt, ... handle 404's on single files in a graceful way.

PR #553, opened 2021-08-19
Runs all DB querries synchronously during a sync task.

dkliban:
closes: #9252


mdellweg (approved):

mdellweg:
Be sure not to merge before pulp/pulpcore#1553

dralley (approved):

pulp_file/app/tasks/synchronizing.py
```diff
@@ -118,5 +122,5 @@ async def run(self):
                     deferred_download=deferred_download,
                 )
                 dc = DeclarativeContent(content=file, d_artifacts=[da])
-                pb.increment()
+                await pb.aincrement()
```
dralley, line 125: We should be doing this less frequently, not for individual file entires

PR #569, opened 2021-09-02
Add ACS path validation

daviddavis:
fixes #9341

### DIFF
--- a/pulp_file/app/serializers.py
+++ b/pulp_file/app/serializers.py
@@ -169,6 +169,13 @@ class FileAlternateContentSourceSerializer(AlternateContentSourceSerializer):
     Serializer for File alternate content source.
     """
 
+    def validate_paths(self, paths):
+        """Validate that paths do not start with /."""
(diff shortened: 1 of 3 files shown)


pulp_file/app/serializers.py
```diff
@@ -169,6 +169,13 @@ class FileAlternateContentSourceSerializer(AlternateContentSourceSerializer):
     Serializer for File alternate content source.
     """
 
+    def validate_paths(self, paths):
+        """Validate that paths do not start with /."""
+        for path in paths:
```
goosemania, line 174: Is it allowed for path to be an empty string? Is considered as a root of the base url they provided?

goosemania (approved):

pulp_file/app/serializers.py
pavelpicka: In pulp_file there should not be an empty path allowed as user must specify at least one manifest file (as a path).

PR #572, opened 2021-09-08
Adopt new dispatch() API for 3.16 compatibility

dralley:
[noissue]


pulp_file/app/viewsets.py
```diff
@@ -161,7 +162,7 @@ def create(self, request):
 
         result = dispatch(
             tasks.publish,
-            [repository_version.repository],
+            exclusive_resources=[repository_version.repository],
```
ipanova, line 165: thinking of this more, since this is a read operation on the repo, does it make sense to keep it in the shared resources?

dralley: That would make sense.  Should we change it in other plugins too?

ipanova: +1

PR #576, opened 2021-09-16
Add ACS functional tests

daviddavis:


daviddavis:
This needs to be rebased against https://github.com/pulp/pulp_file/pull/536.

daviddavis:
The build failure is due to https://pulp.plan.io/issues/9417

pulp_file/tests/functional/api/test_acs.py
```diff
@@ -25,33 +46,123 @@ def setUpClass(cls):
...
+        with self.assertRaises(PulpTaskError) as ctx:
+            monitor_task(sync_response.task)
+        self.assertIn("404", ctx.exception.task.error["description"])
+
+        # create an acs and pull in its remote artifacts
+        acs = self._create_acs()
+        resp = self.file_acs_api.refresh(acs.pulp_href, acs)
+        monitor_task_group(resp.task_group)
+
+        # the sync show now work as the files are being pulled from ACS remote
```
ipanova, line 109: s/show/should

pulp_file/tests/functional/api/test_acs.py
```diff
@@ -25,33 +46,123 @@ def setUpClass(cls):
...
+    def test_acs_sync_with_paths(self):
+        """Test syncing from an ACS using different paths."""
+        repo = self.repo_api.create(gen_repo())
+        self.addCleanup(self.repo_api.delete, repo.pulp_href)
+
+        remote = self.file_remote_api.create(gen_file_remote(FILE_MANIFEST_ONLY_FIXTURE_URL))
+        self.addCleanup(self.file_remote_api.delete, remote.pulp_href)
+
+        acs = self._create_acs(
+            paths=("file/PULP_MANIFEST", "file2/PULP_MANIFEST"),
```
ipanova, line 122: I suggest to assert that after `refresh` 2 hidden repos were created

pulp_file/tests/functional/api/test_acs.py
```diff
@@ -25,33 +46,123 @@ def setUpClass(cls):
...
+        acs = self._create_acs()
+        resp = self.file_acs_api.refresh(acs.pulp_href, acs)
+        monitor_task_group(resp.task_group)
+
+        repository_sync_data = RepositorySyncURL(remote=remote.pulp_href)
+        sync_response = self.repo_api.sync(repo.pulp_href, repository_sync_data)
+        monitor_task(sync_response.task)
+        repo = self.repo_api.read(repo.pulp_href)
+
+        # Create a publication.
```
ipanova, line 152: I was thinking we could use auto-distribution feature here?

pulp_file/tests/functional/api/test_acs.py
daviddavis: How?

pulp_file/tests/functional/api/test_acs.py
daviddavis: This was copy-and-pasted from test_download_content. I have no objection to use auto-distribute but I also don't see an advantage?

ipanova: there is not really any difference just probably less lines of code

pulp_file/tests/functional/api/test_acs.py
ipanova: hm, well, not trough bindings >.>

ipanova (approved):

pulp_file/tests/functional/api/test_acs.py
ipanova: @daviddavis file and file2 repos seemsto have different digests for content so maybe counting artifacts/units ==6 pieces would show that refresh from difference paths work as it should?

daviddavis: The only way to assert the artifact count would be to start from a known state (such as zero) which might be hard or problematic especially in dev setups. 

I think I could assert that that two artifacts (one from each repo) exist after the refresh. I can look into this.

daviddavis:
I'm checking the number of tasks after refresh in the `test_acs_sync_with_paths` test and I updated the `test_serving_acs_content` to use auto-publish/auto-distribute.

ipanova (approved):

PR #588, opened 2021-10-05
Update the pulpcore requirement

daviddavis:
[noissue]

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,1 +1,1 @@
-pulpcore>=3.15.0
+pulpcore>=3.15.0,<3.18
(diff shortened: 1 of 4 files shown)


requirements.txt
```diff
@@ -1 +1 @@
-pulpcore>=3.15.0
+pulpcore>=3.15.0,<3.18
```
dkliban, line 1: do we normally set an upper bound on the main branch?

daviddavis: Oops, I used the wrong base branch. Should be fixed now.

dkliban (approved):

PR #591, opened 2021-10-08
Removed distribution validation which is handled in core.

ipanova:
[noissue]


docs/workflows/publish-host.rst
```diff
@@ -69,7 +69,7 @@ updated automatically when new repository versions are created.
     pulp file repository update --name $REPO_NAME --autopublish
 
     # This configures the distribution to be track the latest repository version for a given repository
-    pulp file distribution update --name $DIST_NAME --repository $REPO_NAME
+    pulp file distribution update --name $DIST_NAME --repository $REPO_NAME --publication ""
```
mdellweg, line 72: I believe, this is what may fail the nightly cli tests. Should the cli attempt to silently unset the publication in that case?

ipanova: I am fixing this in cli as well.

mdellweg: Just merged the CLI change,

PR #598, opened 2021-11-30
Updates GHA configs to latest from plugin_template

dkliban:
This also accounts for a change in pulpcore branch name to main.

[noissue]


bmbouter (approved):
Thanks!

PR #636, opened 2022-01-20
Add new Roles

gerrod3:
Still lots to work on. For the most part I think the file plugin won't have any other roles outside of the three defaults we have been using: creator, owner, & viewer. We still need to answer the question of content ownership, but I think that needs to be addressed in pulpcore as it affects all plugins.

TODO:

- [x] Figure out the correct access policy for repository versions
- [x] Add new pytest style tests
- [x] Find and add changelog issue


pulp_file/app/viewsets.py
```diff
@@ -73,6 +73,23 @@ class FileContentViewSet(SingleArtifactContentUploadViewSet):
     serializer_class = FileContentSerializer
     filterset_class = FileContentFilter
 
+    DEFAULT_ACCESS_POLICY = {
+        # I think this doesn't do anything different from the default permission class
```
mdellweg, line 77: It does. The default is restricted to admins.

pulp_file/app/viewsets.py
```diff
@@ -84,9 +101,79 @@ class FileRepositoryViewSet(RepositoryViewSet, ModifyRepositoryActionMixin):
...
+                "effect": "allow",
+                "condition": "has_model_or_obj_perms:file.view_filerepository",
+            },
+            {
+                "action": ["destroy"],
+                "principal": "authenticated",
+                "effect": "allow",
+                "condition": [
+                    "has_model_or_obj_perms:file.delete_filerepository",
+                    "has_model_or_obj_perms:file.view_filerepository", # is this needed with queryset_filtering?
```
mdellweg, line 131: This is debatable at best. Do you really need to see something if you are allowed to delete it?
I would say for the sake of simplicity we can not check "view" here and assume that every role that has delete also contains view.

pulp_file/app/viewsets.py
```diff
@@ -125,6 +212,10 @@ class FileRepositoryVersionViewSet(RepositoryVersionViewSet):
     """
 
     parent_viewset = FileRepositoryViewSet
+    # Hmm, not sure what the access policy should look like
+    # Taking a look at https://github.com/pulp/pulp_container/pull/403/ which is using
+    # https://github.com/pulp/pulpcore/pull/1800, is different from how I would do it
+    # and I don't understand the difference
```
mdellweg, line 218: pulpcore/#1800 is helping to delegate the permissions to the repository so you never have permissions on the versions directly.

mdellweg (commented):
I know this is not finished.
I was missing some creation hooks and the nested viewsets to manage roles.

pulp_file/app/viewsets.py
gerrod3: True, but my question was that all items should be queryset filtered by the `queryset_filtering_required_permission` which is usually always the view permission, so is this extra condition even necessary? I think even without this condition you still couldn't delete items you couldn't see.

pulp_file/tests/functional/utils.py
```diff
@@ -152,3 +159,46 @@ def gen_artifact(url=FILE_URL, file=None):
             return ArtifactsApi(core_client).create(file=temp_file.name).to_dict()
 
     return ArtifactsApi(core_client).create(file=file).to_dict()
+
+
+# This should probably go into pulp_smash as the default gen_user function
+def gen_user_rest(cfg=None, model_roles=None, object_roles=None, **kwargs):
```
mdellweg, line 165: How do you think about making this a factory fixture?

gerrod3: Probably a good idea. I don't really understand the difference between having a function as a fixture factory vs just a normal utility function.

mdellweg: The fixture factory can keep track of the generated items and clean them up automatically.

pulp_file/app/models.py
```diff
@@ -42,31 +43,41 @@ class Meta:
...
 
 
-class FileRemote(Remote):
+class FileRemote(Remote, AutoAddObjPermsMixin):
     """
     Remote for "file" content.
     """
 
     TYPE = "file"
+    ACCESS_POLICY_VIEWSET_NAME = "remotes/file/file"
```
mdellweg, line 52: Didn't we render this superfluous?

gerrod3: Yes, I'll update it.

pulp_file/tests/functional/api/test_rbac.py
```diff
@@ -0,0 +1,201 @@
...
+def _try_action(user, action_api, outcome, *args, **kwargs):
+    try:
+        with user:
+            response = action_api(*args, **kwargs)
+        if isinstance(response, AsyncOperationResponse):
+            response = monitor_task(response.task)
+    except ApiException as e:
+        assert e.status == outcome, f"{e}"
+    else:
+        assert outcome == 200, "User performed action they shouldn't been able to"
```
mdellweg, line 43: Shouldn't we check `response.status` here? Sometimes we want to see a 203...

pulp_file/tests/functional/api/test_rbac.py
```diff
@@ -0,0 +1,201 @@
...
+        creator_roles = [f"file.{role}_creator" for role in role_names]
+        alice = gen_user(model_roles=viewer_roles)
+        bob = gen_user(model_roles=creator_roles)
+        charlie = gen_user()
+        return alice, bob, charlie
+
+    return _gen_users
+
+
+def _try_action(user, action_api, outcome, *args, **kwargs):
```
mdellweg, line 34: Do you think this should be part of the Pulp RBAC testing framework (PRTF)?

pulp_file/tests/functional/api/test_rbac.py
```diff
@@ -0,0 +1,201 @@
...
+    # Delete testing
+    _try_action(alice, file_repo_api_client.delete, 403, repo.pulp_href)
+    _try_action(charlie, file_repo_api_client.delete, 404, repo.pulp_href)
+    _try_action(bob, file_repo_api_client.delete, 200, repo.pulp_href)
+
+
+@pytest.mark.parallel
+def test_role_management(gen_users, file_repo_api_client, gen_object_with_cleanup):
+    """Check that role management apis."""
+    alice, bob, charlie = gen_users("filerepository")
```
mdellweg, line 83: We could use marks to parameterize `alice` `bob` and `charly` fixtures. WDYT?

By the way, in pulp_container we call them `user_creator`, `user_reader` and `user_helpless`.

gerrod3: Sounds interesting, I'll try it out.

pulp_file/tests/functional/api/test_rbac.py
gerrod3: Yeah, that would make it more versatile.

.github/workflows/scripts/func_test_script.sh
```diff
@@ -21,4 +21,4 @@ fi
 
 pip install -r ../pulpcore/functest_requirements.txt
 
-pytest -v -r sx --color=yes --pyargs pulpcore.tests.functional
+pytest -v -r sx --color=yes --pyargs pulpcore.tests.functional -m "not nightly"
```
gerrod3, line 24: Maybe, I should also update the one above for pulp_file to not run the nightly marked tests? Also, I think this is the area for creating a parallel test runner as well. 

The remove_plugin tests in pulpcore will fail after this is merged (it is marked nightly), so I'll have to create a PR to fix that as well.

mdellweg: This file is not covered by the plugin_template, right? Then it needs manual curation. :+1:

pulp_file/app/viewsets.py
```diff
@@ -84,9 +95,98 @@ class FileRepositoryViewSet(RepositoryViewSet, ModifyRepositoryActionMixin):
...
+                "principal": "authenticated",
+                "effect": "allow",
+            },
+            {
+                "action": ["create"],
+                "principal": "authenticated",
+                "effect": "allow",
+                "condition": [
+                    "has_model_perms:file.add_filerepository",
+                    "has_remote_param_model_or_obj_perms:file.view_fileremote",
```
mdellweg, line 113: So this is to prevent you consuming a remote you may not see? Is this "also" prevented by queryset scoping?

pulp_file/app/viewsets.py
```diff
@@ -126,8 +226,37 @@ class FileRepositoryVersionViewSet(RepositoryVersionViewSet):
...
+                "action": ["destroy"],
+                "principal": "authenticated",
+                "effect": "allow",
+                "condition": [
+                    "has_repository_model_or_obj_perms:file.delete_filerepository",
+                    "has_repository_model_or_obj_perms:file.view_filerepository",
+                ],
+            },
+            {
+                "action": ["repair"],  # should repair get its own permission?
```
mdellweg, line 247: Good question. Consider: The access policy can be adjusted, but only using permissions we already provide. Also the idea of roles is to combine sets of permissions, so permissions should be considered cheap.

.github/workflows/scripts/func_test_script.sh
bmbouter: Yeah I think it also needs the parallel script line too. These are the two lines from the pulpcore diff for example:

```
.github/workflows/scripts/script.sh:        pytest -v -r sx --color=yes --suppress-no-test-exit-code --pyargs pulpcore.tests.functional -m parallel -n 8
.github/workflows/scripts/script.sh:        pytest -v -r sx --color=yes --pyargs pulpcore.tests.functional -m "not parallel"
```

CHANGES/626.feature
```diff
@@ -0,0 +1,5 @@
+Added Role Based Access Control for each endpoint.
```
bmbouter, line 1: Are these intended as bullets or one paragraph? If one paragraph can you line wrap beyond the `.`, or if its bullets maybe add the `    * `?

bmbouter: The content looks good to go tho. :+1:

gerrod3: I'll add the bullets.

pulp_file/app/viewsets.py
gerrod3: No, in this case no queryset scoping is done on the file remote when creating a repository. We only do queryset scoping on the `get_queryset` method of the viewset, which will be on that view's object (repository in this case). This is why we need this condition check to make sure you can't make a repository with a remote you don't have permission for.

CHANGES/626.feature
```diff
@@ -0,0 +1,3 @@
...
+* New default roles (creator, owner, viewer) have been added for ``FileRepository``, ``FileRemote``, ``FileDistribution``, ``FilePublication``, and ``FileAlternateContentSource``.
```
bmbouter, line 2: nitpick but can this line wrap to 100 chars

bmbouter:
This looks really great! I'm so excited for it.

pulp_file/app/viewsets.py
mdellweg: This is still an outstanding question. If we intend to merge this without an answer, we need to capture the discussion elsewhere.

ipanova: @gerrod3 @bmbouter ^

ipanova: we had similar debate when updating container policy. We ended up using sync permission. Repair is based on remote content artifact which is created only during sync operation :)

pulp_file/app/viewsets.py
```diff
@@ -73,8 +74,18 @@ class FileContentViewSet(SingleArtifactContentUploadViewSet):
     serializer_class = FileContentSerializer
     filterset_class = FileContentFilter
 
+    DEFAULT_ACCESS_POLICY = {
+        "statements": [
+            {
+                "action": ["list", "retrieve", "create"],
```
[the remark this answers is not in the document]
gerrod3: It is purposely without queryset scoping. We have not decided on how permissions for content uploaded to Pulp should be handled.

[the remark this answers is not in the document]
ipanova: what about content that was mirrored into pulp?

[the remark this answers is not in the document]
gerrod3: Anybody that is authenticated could view that content and add it to their own repositories. Since content is dedupped it is hard to give individual permissions.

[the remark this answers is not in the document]
ipanova: I understand the complexity and that's why we have not really resolved the content RBAC yet. The main question I am raising here is how we approach content access in each plugin *right now*. If you look at the container plugin one can view only that content he has read access to the repos to. If you look at file plugin then it is enough to be authed. 

Do we want to keep consistency between plugins? Do we add similar approach container has to the file? Do we remove the approach we have in container today so it matches what file has? Or we just keep adding stuff without syncing the logic between plugins?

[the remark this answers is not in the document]
ipanova: @mdellweg @bmbouter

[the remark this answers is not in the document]
mdellweg: I would absolutely vote for consistency. Preferably by going the container way.

pulp_file/app/viewsets.py
mdellweg: With the roles, it should not be a big burden for day to day operation to have an additional permission.

pulp_file/app/viewsets.py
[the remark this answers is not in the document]
bmbouter: I *think* I'm understanding the convo, but apologies if I'm off base here. The RBAC initiative is so big we have to do it in pieces. For better or worse (my understanding) of the current plan is that we're going to rbac everything except content. To do the content part right I think we'll end up having to have some design calls and discussions (basically a whole new phase of RBAC). So if that's the case, my believe is we should accept this PR and add the content stuff later. Wdyt?

gerrod3:
@bmbouter @mdellweg Can you do a final review so we can get this merged?

mdellweg (approved):
Can you please make sure that we track the Question of securing the access to content in a ticket?

PR #665, opened 2022-03-25
Revert "Add all tests to the package for easy * importing"

bmbouter:
This reverts commit 2cabcf2ba5aeb73d4ba08760cf35627ca1430ddf.

[noissue]


bmbouter:
We don't want to start this requirement if we don't have to. Now that we're going to run the functional tests directly with pytest, we don't need this.

mdellweg (approved):

PR #675, opened 2022-04-07
Add missing test.

ipanova:
[noissue]

### DIFF
--- a/pulp_file/tests/functional/api/test_crud_remotes.py
+++ b/pulp_file/tests/functional/api/test_crud_remotes.py
@@ -39,6 +39,7 @@ class CRUDRemotesTestCase(unittest.TestCase):
         self._partially_update()
         self._fully_update()
         self._delete()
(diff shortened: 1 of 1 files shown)


pulp_file/tests/functional/api/test_crud_remotes.py
```diff
@@ -107,7 +108,7 @@ def _delete(self):
         with self.assertRaises(ApiException):
             self.remote_api.read(self.remote.pulp_href)
 
-    def test_negative_create_file_remote_with_invalid_parameter(self):
+    def _test_negative_create_file_remote_with_invalid_parameter(self):
```
mdellweg, line 111: I'm wondering whether this test was left separate for the reason that it wasn't involved in the remotes lifecycle of the other tests.

ipanova: hm, looking from this angle, you might be right

PR #681, opened 2022-04-12
Add no-leftovers check for File ACS and FileRemote

bmbouter:
[noissue]


bmbouter:
This checks for the pulp_file specific objects that pulpcore couldn't check on its own in [this PR](https://github.com/pulp/pulp-smash/pull/1285).

bmbouter:
@mdellweg can you review this one too plz? Thanks!

PR #686, opened 2022-04-18
Makes pulp_file *not* a pytest plugin

bmbouter:
We've consolidated all the pulp_file tests into the pulp_file repo, so
there is no reason to have pulp_file provide fixtures externally
anymore.

[noissue]

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -31,8 +31,5 @@ setup(
         "Programming Language :: Python :: 3.8",
         "Programming Language :: Python :: 3.9",
     ),
-    entry_points={
-        "pulpcore.plugin": ["pulp_file = pulp_file:default_app_config"],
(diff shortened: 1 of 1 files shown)


bmbouter:
This will be undrafted after https://github.com/pulp/pulpcore/pull/2535 and https://github.com/pulp/pulp_file/pull/685 are merged.

mdellweg:
Let me reiterate: There is an easy way to use the fixtures without making it a pytest plugin. And this is the way is see us consume pulpcore fixtures in plugin CI.

https://docs.pytest.org/en/7.1.x/how-to/fixtures.html#using-fixtures-from-other-projects

mdellweg (approved):

PR #687, opened 2022-04-20
Have pytest load fixtures from pulpcore

bmbouter:
Required PR: https://github.com/pulp/pulpcore/pull/2562

[noissue]

### DIFF
--- a/pulp_file/tests/functional/conftest.py
+++ b/pulp_file/tests/functional/conftest.py
@@ -24,6 +24,9 @@ from pulp_file.tests.functional.utils import gen_file_client
 _logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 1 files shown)


pulp_file/tests/functional/conftest.py
```diff
@@ -24,6 +24,9 @@
 _logger = logging.getLogger(__name__)
 
 
+pytest_plugins = "pulpcore.tests.functional.conftest"
```
bmbouter, line 27: I believe the CI will fail here like it does locally. This is definitly being parsed, but it can't find the fixtures from pulpcore. I think pytest expects this to be a package not a module. Should I look into this more or go back to the other pattern?

mdellweg (approved):

dkliban (approved):

PR #689, opened 2022-04-20
Add test for bad network conditions

gerrod3:
[noissue]

### DIFF
--- a/pulp_file/tests/functional/api/test_bad_sync.py
+++ b/pulp_file/tests/functional/api/test_bad_sync.py
@@ -1,0 +1,70 @@
+import pytest
+import uuid
+
+from collections import defaultdict
+from pulp_smash.pulp3.bindings import monitor_task
+
+from pulpcore.client.pulp_file import (
+    RepositorySyncURL,
+)
+
+
+@pytest.fixture
+def perform_sync(file_repo, file_repo_api_client, file_remote_api_client, gen_object_with_cleanup):
+    def _perform_sync(url, policy="immediate"):
+        remote_data = {
+            "url": str(url),
+            "policy": policy,
+            "name": str(uuid.uuid4()),
+        }
+        remote = gen_object_with_cleanup(file_remote_api_client, remote_data)
+
+        body = RepositorySyncURL(remote=remote.pulp_href)
+        monitor_task(file_repo_api_client.sync(file_repo.pulp_href, body).task)
+        return file_repo
+
+    yield _perform_sync
+
+
+@pytest.mark.parallel
+def test_bad_response_retry(bad_response_fixture_server, large_manifest_path, perform_sync):
+    """
+    Test downloader retrying after network failure during sync.
+    """
+    requests_record = bad_response_fixture_server.requests_record
+    url = bad_response_fixture_server.make_url(large_manifest_path)
+
+    perform_sync(url)
+
+    # 1 for PULP_MANIFEST, and 4 for 1.iso
+    assert len(requests_record) == 5
+    assert "PULP_MANIFEST" in requests_record[0].raw_path
+    for i in range(1, 5):
+        assert "1.iso" in requests_record[i].raw_path
+
+
+@pytest.mark.parallel
+def test_bad_response_retry_multiple_files(
+    bad_response_fixture_server,
+    basic_manifest_path,
+    perform_sync,
+):
+    """
+    Test multiple file downloaders retrying after network failure during sync.
+    """
+    requests_record = bad_response_fixture_server.requests_record
+    url = bad_response_fixture_server.make_url(basic_manifest_path)
+
+    perform_sync(url)
+
+    # 1 for PULP_MANIFEST, and 4 each for 1.iso, 2.iso, 3.iso
+    assert len(requests_record) == 13
+    records = defaultdict(int)
+    for r in requests_record:
+        filename = r.raw_path.split("/")[-1]
+        records[filename] += 1
+
+    assert records["PULP_MANIFEST"] == 1
+    assert records["1.iso"] == 4
+    assert records["2.iso"] == 4
+    assert records["3.iso"] == 4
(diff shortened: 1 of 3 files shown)


pulp_file/tests/functional/conftest.py
```diff
@@ -159,3 +161,31 @@ def _file_fixture_gen_remote_client_cert_req(*, fixture_name, policy, **kwargs):
         return gen_object_with_cleanup(file_remote_api_client, kwargs)
 
     yield _file_fixture_gen_remote_client_cert_req
+
+
+@pytest.fixture
+def invalid_file_fixture(gen_threaded_aiohttp_server, file_fixtures_root):
```
dralley, line 167: This is great.  Instead of calling it `invalid_file_fixture`, can we call it something more descriptive?  And add some descriptive comments.

My initial thought was that this should work fine because you were reading the full amount of bytes of the file, and then I noticed the missing zero at the end of `1048576`.

It might help to move the expected size and the actual size up to the top of the method as constants, so that you can see them side by side.

Also, it would be great to add another fixture which changes one byte of the file while leaving the length intact.

dralley:
Needs rebase

pulp_file/tests/functional/conftest.py
```diff
@@ -159,3 +161,31 @@ def _file_fixture_gen_remote_client_cert_req(*, fixture_name, policy, **kwargs):
...
+    record = []
+    num_requests = 0
+    path = file_fixtures_root / "large"
+
+    async def handler(request):
+        nonlocal num_requests
+        record.append(request)
+        relative_path = request.raw_path[1:]  # Strip off leading "/"
+        # Max retries is 3. So on fourth request, send full data
```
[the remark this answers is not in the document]
gerrod3: This fixture is only for the 'large' file fixture which only has one file to download. `num_requests` is only incremented when `1.iso` is requested (since it is the only content being requested). I can refactor this fixture later if we find more network problems we need to simulate and test.

pulp_file/tests/functional/conftest.py
gerrod3: I renamed it to `bad_response_fixture_server` and added a description for each response it sends.

pulp_file/tests/functional/conftest.py
```diff
@@ -192,3 +204,59 @@ def _file_fixture_gen_file_repo(**kwargs):
...
+            record.append(request)
+            relative_path = request.raw_path[1:]  # Strip off leading "/"
+            file_path = Path(fixtures_root) / Path(relative_path)
+            # Max retries is 3. So on fourth request, send full data
+            num_requests[relative_path] += 1
+            if "PULP_MANIFEST" in relative_path or num_requests[relative_path] % 4 == 0:
+                return web.FileResponse(file_path)
+
+            # On third request send 429 error, TooManyRequests
+            if num_requests[relative_path] == 3:
```
mdellweg, line 235:
```suggestion
            if num_requests[relative_path] % 4 == 3:
```

pulp_file/tests/functional/conftest.py
```diff
@@ -192,3 +204,59 @@ def _file_fixture_gen_file_repo(**kwargs):
...
+
+            size = file_path.stat().st_size
+            response = web.StreamResponse(headers={"content-length": f"{size}"})
+            await response.prepare(request)
+            async with aiofiles.open(file_path, "rb") as f:
+                # Send only partial content causing aiohttp.ClientPayloadError if request num == 1
+                chunk = await f.read(size // 2)
+                await response.write(chunk)
+                # Send last chunk with modified last byte if request num == 2
+                if num_requests[relative_path] == 2:
```
mdellweg, line 246:
```suggestion
                if num_requests[relative_path] % 4 == 2:
```

pulp_file/tests/functional/conftest.py
```diff
@@ -192,3 +204,59 @@ def _file_fixture_gen_file_repo(**kwargs):
...
+                    chunk2 = await f.read()
+                    await response.write(chunk2[:-1])
+                    await response.write(bytes([chunk2[-1] + 1]))
```
mdellweg, line 249: Maybe just flip a bit to be safe:
```suggestion
                    await response.write(bytes([chunk2[-1] ^ 1]))
```

mdellweg (commented):
I like the concept to corrupt the data in various ways.

mdellweg (approved):

pulp_file/tests/functional/api/test_bad_sync.py
```diff
@@ -0,0 +1,67 @@
...
+):
+    """
+    Test multiple file downloaders retrying after network failure during sync.
+    """
+    requests_record = bad_response_fixture_server.requests_record
+    url = bad_response_fixture_server.make_url(basic_manifest_path)
+
+    perform_sync(url)
+
+    # 1 for PULP_MANIFEST, and 4 for 1.iso, 2.iso, 3.iso
```
dralley, line 60: 4 "each"

pulp_file/tests/functional/api/test_bad_sync.py
```diff
@@ -0,0 +1,67 @@
...
+    Test multiple file downloaders retrying after network failure during sync.
+    """
+    requests_record = bad_response_fixture_server.requests_record
+    url = bad_response_fixture_server.make_url(basic_manifest_path)
+
+    perform_sync(url)
+
+    # 1 for PULP_MANIFEST, and 4 each for 1.iso, 2.iso, 3.iso
+    assert len(requests_record) == 13
+    assert "PULP_MANIFEST" in requests_record[0].raw_path
```
dralley, line 62: How would you feel about doing it this way?  (filename split might need to be slightly different, haven't tested it)

```Python
records = defaultdict(int)
for r in requests_record:
    filename = r.raw_path.split("/")[-1]
    records[filename] += 1
    
assert records["PULP_MANIFEST"] == 1
assert records["1.iso"] == 4
assert records["2.iso"] == 4
assert records["3.iso"] == 4
```

dralley (approved):

Issue #691, opened 2022-04-22
Generate file repositories that are served by pytest server fixture


dkliban:
The pytest server fixtures currently use files that are comitted to the pulp-file repository. The same files are used by 6 different tests. This requires removing orphans before each of those tests is run. If the files were generated for each test, the orphan cleanup would not be needed. The tests could then be parallelized also.

PR #692, opened 2022-04-22
Generates random file repositories for pytest server fixture

dkliban:
closes: #691


mdellweg (approved):

dralley (approved):
Nice

pulp_file/tests/functional/conftest.py
```diff
@@ -81,9 +83,30 @@ def file_remote_api_client(file_client):
     return RemotesFileApi(file_client)
 
 
-@pytest.fixture(scope="session")
-def file_fixtures_root():
-    return Path(__file__).parent / "fixtures"
+def generate_iso(name, size=1024):
```
bmbouter, line 86: Typically conftest.py doesn't contain utilities but only pytest things. If these stay as utils they probably should move to the utils.py. If they stay here I think these should become actual pytest fixtures.

mdellweg: Private functions should always be ok: `_generate_iso`.

PR #698, opened 2022-04-25
Changes utility functions into factory fixtures

dkliban:
[noissue]

### DIFF
--- a/pulp_file/tests/functional/conftest.py
+++ b/pulp_file/tests/functional/conftest.py
@@ -86,25 +86,33 @@ def file_remote_api_client(file_client):
     return RemotesFileApi(file_client)
 
 
-def generate_iso(name, size=1024):
-    with open(name, "wb") as fout:
-        contents = os.urandom(size)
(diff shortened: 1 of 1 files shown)


bmbouter (approved):
thank you!

pulp_file/tests/functional/conftest.py
```diff
@@ -86,25 +86,33 @@ def file_remote_api_client(file_client):
...
 
 
-def generate_manifest(name, file_list):
-    with open(name, "wt") as fout:
-        for file in file_list:
-            fout.write("{},{},{}\n".format(file["name"], file["digest"], file["size"]))
-        fout.flush()
-    return name
+@pytest.fixture()
+def generate_manifest():
```
mdellweg, line 103: Where's the benefit here? I feel like we are overdoing it. The fixture cleanup is handled by the `file_fixture_root` using `tmpdir`. OTOH using this fixture elsewhere is completely lacking any cleanup.

PR #699, opened 2022-04-27
Update CI files from main branch

pulpbot:
[noissue]


gerrod3 (approved):

gerrod3:
Do we care about the clarity of our git history? These automatic updates only affect the CI files, but they do clog up the commit log.

bmbouter:
I've had to apply more merges than I'd like, but in normal conditions the idea is to only have it happen automatically once a week.

PR #709, opened 2022-05-05
Adds "basic_manifest_path" fixture

dkliban:
This fixture replaces the "file_fixture_basic". The new fixture can be used by
tests directly. This introduces a new pattern for writing file fixtures that will
allow each test to specify which specific randomly generated fixture it should
use. In the future we will add fixtures that represent other file fixtures that
are present in fixtures.pulpproject.org.

[noissue]

### DIFF
--- a/pulp_file/tests/functional/conftest.py
+++ b/pulp_file/tests/functional/conftest.py
@@ -83,11 +83,17 @@ def file_remote_api_client(file_client):
 
 @pytest.fixture()
 def file_fixtures_root(tmpdir):
-    file1 = generate_iso(tmpdir.join("1.iso"))
-    file2 = generate_iso(tmpdir.join("2.iso"))
-    file3 = generate_iso(tmpdir.join("3.iso"))
-    generate_manifest(tmpdir.join("PULP_MANIFEST"), [file1, file2, file3])
     return Path(tmpdir)
+
+
+@pytest.fixture()
+def basic_manifest_path(file_fixtures_root):
+    file_fixtures_root.joinpath("basic").mkdir()
+    file1 = generate_iso(file_fixtures_root.joinpath("basic/1.iso"))
+    file2 = generate_iso(file_fixtures_root.joinpath("basic/2.iso"))
+    file3 = generate_iso(file_fixtures_root.joinpath("basic/3.iso"))
+    generate_manifest(file_fixtures_root.joinpath("basic/PULP_MANIFEST"), [file1, file2, file3])
+    return "/basic/PULP_MANIFEST"
 
 
 @pytest.fixture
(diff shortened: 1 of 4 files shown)


pulp_file/tests/functional/conftest.py
```diff
@@ -3,6 +3,8 @@
 
 from pathlib import Path
 
+import py
```
bmbouter, line 6: Is this supposed to be here?

dkliban: I use it on line 93. `tmpdir` fixture is of type `py.path.local`. The utility methods for generating the files and the manifest expect the path passed in to them to be of type `local`.

The web server fixture expects the path to be of type `pathlib.Path`. This means that `file_fixtures_root` needs to return a `pathlib.Path`. This means that the fixtures that use `file_fixtures_root` and the utility methods for generating files need to convert a `Path` to `local`. 

Another option is to rewrite the utility methods to use `Path`.

pulp_file/tests/functional/conftest.py
```diff
@@ -83,13 +85,18 @@ def file_remote_api_client(file_client):
...
-    file1 = generate_iso(tmpdir.join("1.iso"))
-    file2 = generate_iso(tmpdir.join("2.iso"))
-    file3 = generate_iso(tmpdir.join("3.iso"))
-    generate_manifest(tmpdir.join("PULP_MANIFEST"), [file1, file2, file3])
     return Path(tmpdir)
 
 
+@pytest.fixture()
+def file_fixture_basic(file_fixtures_root):
+    root = py.path.local(file_fixtures_root)
```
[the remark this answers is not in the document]
dkliban: https://github.com/pytest-dev/py

mdellweg:
I need to think about this new pattern. I can see where you are coming from. But, would it be better to introduce the fixture parameterization by pytest.mark? What if a test requests two such fixtures? Should we move them to subdirectories?

pulp_file/tests/functional/conftest.py
bmbouter: I think having the utilities use `Path` natively would be nice. Is there a downside to that?

dkliban: I updated the code to use Path.

mdellweg (approved):

pulp_file/tests/functional/conftest.py
```diff
@@ -83,13 +83,18 @@ def file_remote_api_client(file_client):
...
     return Path(tmpdir)
 
 
+@pytest.fixture()
+def file_fixture_basic(file_fixtures_root):
+    file_fixtures_root.joinpath("basic").mkdir()
+    file1 = generate_iso(file_fixtures_root.joinpath("basic/1.iso"))
+    file2 = generate_iso(file_fixtures_root.joinpath("basic/2.iso"))
+    file3 = generate_iso(file_fixtures_root.joinpath("basic/3.iso"))
+    generate_manifest(file_fixtures_root.joinpath("basic/PULP_MANIFEST"), [file1, file2, file3])
```
bmbouter, line 95: It would be nice if this fixture returned the full path to the manifest so here that would be `"basic/PULP_MANIFEST"`. I think it's better if the test doesn't have to assume things about the layout and instead be told the partial path to the manifest. For example cases like where instead of `PULP_MANIFEST` it's `SOME_OTHER_NAME`. So then the `make_url(fixture_name)` throughout. Maybe rename `fixture_name` to `relative_path_to_manifest` or something.

bmbouter (approved):
Looks good to me!

PR #712, opened 2022-05-06
Port test_label tests

bmbouter:
[noissue]


dkliban (approved):
Thanks for porting these.

PR #715, opened 2022-05-13
Use certguard fixtures

mdellweg:
[noissue]


.github/workflows/scripts/install.sh
```diff
@@ -20,7 +20,8 @@ if [[ "$TEST" = "docs" || "$TEST" = "publish" ]]; then
   pip install -r doc_requirements.txt
 fi
 
-pip install -e ../pulpcore
+# TODO reflect this in the plugin-template
+pip install -e ../pulpcore -e ../pulp-certguard
```
mdellweg, line 24: This is why it's draft now.

mdellweg: https://github.com/pulp/plugin_template/pull/643

bmbouter (approved):
Looks great!

PR #727, opened 2022-05-27
Port the test_crud_remotes to pytest

bmbouter:
[noissue]

### DIFF
--- a/pulp_file/tests/functional/api/test_crud_remotes.py
+++ b/pulp_file/tests/functional/api/test_crud_remotes.py
@@ -1,204 +1,98 @@
 """Tests that CRUD file remotes."""
 import json
-from random import choice
-import unittest
+import uuid
 
-from pulp_smash import utils
+import pytest
 
 from pulp_smash.pulp3.bindings import monitor_task
-from pulp_smash.pulp3.constants import ON_DEMAND_DOWNLOAD_POLICIES
 
-from pulp_file.tests.functional.constants import (
-    FILE_FIXTURE_MANIFEST_URL,
-    FILE2_FIXTURE_MANIFEST_URL,
-)
-from pulp_file.tests.functional.utils import (
-    gen_file_client,
-    gen_file_remote,
-)
-
-from pulpcore.client.pulp_file import RemotesFileApi
 from pulpcore.client.pulp_file.exceptions import ApiException
 
 
-class CRUDRemotesTestCase(unittest.TestCase):
-    """CRUD remotes."""
+@pytest.mark.parallel
+def test_remote_crud_workflow(file_remote_api_client, gen_object_with_cleanup):
+    remote_data = {"name": str(uuid.uuid4()), "url": "http://example.com"}
+    remote = gen_object_with_cleanup(file_remote_api_client, remote_data)
+    assert remote.url == remote_data["url"]
+    assert remote.name == remote_data["name"]
 
-    @classmethod
-    def setUpClass(cls):
-        cls.remote_api = RemotesFileApi(gen_file_client())
+    with pytest.raises(ApiException) as exc:
+        gen_object_with_cleanup(file_remote_api_client, remote_data)
+    assert exc.value.status == 400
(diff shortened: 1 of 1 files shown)


pulp_file/tests/functional/api/test_crud_remotes.py
```diff
@@ -1,204 +1,105 @@
...
-    attrs.update(
-        {
-            "password": utils.uuid4(),
-            "username": utils.uuid4(),
-            "policy": choice(ON_DEMAND_DOWNLOAD_POLICIES),
-        }
+@pytest.mark.parallel
+def test_remote_crud_workflow(file_remote_api_client):
+    remote_data = {"name": str(uuid.uuid4()), "url": "http://example.com"}
+    remote = file_remote_api_client.create(remote_data)
```
mdellweg, line 15: This should use create with auto cleanup.

pulp_file/tests/functional/api/test_crud_remotes.py
```diff
@@ -1,204 +1,105 @@
...
+    delete_reponse = file_remote_api_client.delete(remote.pulp_href)
+    monitor_task(delete_reponse.task)
+
+    with pytest.raises(ApiException) as exc:
+        file_remote_api_client.read(remote.pulp_href)
+    assert exc.value.status == 404
+
+
+@pytest.mark.parallel
+def test_negative_create_file_remote_with_invalid_parameter(file_remote_api_client):
```
mdellweg, line 51:
```suggestion
def test_create_file_remote_with_invalid_parameter(file_remote_api_client):
```
I do not see the value in that piece of the name. The fact that it fails is well encoded (and readable) in the assertions.

pulp_file/tests/functional/api/test_crud_remotes.py
```diff
@@ -1,204 +1,105 @@
...
+@pytest.mark.parallel
+def test_negative_create_file_remote_without_url(file_remote_api_client):
+    with pytest.raises(ApiException) as exc:
+        file_remote_api_client.create({"name": str(uuid.uuid4())})
+    assert exc.value.status == 400
+    assert json.loads(exc.value.body) == {"url": ["This field is required."]}
+
+
+@pytest.mark.parallel
+def test_verify_default_remote_policy_immediate(file_remote_api_client, gen_object_with_cleanup):
```
mdellweg, line 73:
```suggestion
def test_default_remote_policy_is_immediate(file_remote_api_client, gen_object_with_cleanup):
```

`test_verify` sounds like we'd test some verification.

pulp_file/tests/functional/api/test_crud_remotes.py
bmbouter: Usually it would, but this test wants to also assert the ability to delete at the end so it's done this way here.

bmbouter:
@mdellweg I pushed the changes, please re-review. Thanks!

pulp_file/tests/functional/api/test_crud_remotes.py
bmbouter: I agree, basically always use `gen_object_with_cleanup`. I've replaced all `create()` usage in this PR with `gen_object_with_cleanup`.

pulp_file/tests/functional/api/test_crud_remotes.py
```diff
@@ -1,204 +1,98 @@
...
+@pytest.mark.parallel
+def test_create_file_remote_with_invalid_parameter(file_remote_api_client, gen_object_with_cleanup):
+    unexpected_field_remote_data = {
+        "name": str(uuid.uuid4()),
+        "url": "http://example.com",
+        "foo": "bar",
+    }
+
+    with pytest.raises(ApiException) as exc:
+        gen_object_with_cleanup(file_remote_api_client, unexpected_field_remote_data)
```
mdellweg, line 52: Now I start to thinking, maybe we can even hook into the `file_remote_api_client` fixture and monkey patch it's create to add cleanup without having to bother in test code.

Issue #729, opened 2022-06-02
Restrict non-admin users to only oneshot uploads


gerrod3:
The new RBAC policy on content makes content uploads as a non-admin useless as they won't be able to see the orphaned content they just uploaded. Restrict their upload ability to only oneshots (specify a repository).

PR #730, opened 2022-06-02
Add new condition on file uploads

gerrod3:
fixes: #729
Required PR: https://github.com/pulp/pulpcore/pull/2797

### DIFF
--- a/pulp_file/app/viewsets.py
+++ b/pulp_file/app/viewsets.py
@@ -77,9 +77,18 @@ class FileContentViewSet(SingleArtifactContentUploadViewSet):
     DEFAULT_ACCESS_POLICY = {
         "statements": [
             {
-                "action": ["list", "retrieve", "create"],
+                "action": ["list", "retrieve"],
                 "principal": "authenticated",
                 "effect": "allow",
+            },
+            {
+                "action": ["create"],
+                "principal": "authenticated",
+                "effect": "allow",
+                "condition": [
+                    "has_required_repo_perms_on_upload:file.modify_filerepository",
+                    "has_required_repo_perms_on_upload:file.view_filerepository",
+                ],
             },
         ],
         "queryset_scoping": {"function": "scope_queryset"},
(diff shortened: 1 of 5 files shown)


pulp_file/app/viewsets.py
```diff
@@ -77,9 +77,15 @@ class FileContentViewSet(SingleArtifactContentUploadViewSet):
...
-                "action": ["list", "retrieve", "create"],
+                "action": ["list", "retrieve"],
+                "principal": "authenticated",
+                "effect": "allow",
+            },
+            {
+                "action": ["create"],
                 "principal": "authenticated",
                 "effect": "allow",
+                "condition": "has_required_repo_perms_on_upload:file.view_filerepository",
```
mdellweg, line 88: View repository is IMHO not sufficient. This will create a repo version, so we need more Perm.

ipanova: + `has_model_or_obj_perms:file.modify_filerepository`

pulp_file/tests/functional/utils.py
```diff
@@ -121,6 +121,7 @@ def gen_artifact(url=FILE_URL, file=None):
         response = requests.get(url)
         with NamedTemporaryFile() as temp_file:
             temp_file.write(response.content)
+            temp_file.flush()
```
ipanova, line 124: :+1:

pulp_file/app/viewsets.py
gerrod3: Just modify or do we need a new special upload permission?

pulp_file/app/viewsets.py
```diff
@@ -77,9 +77,15 @@ class FileContentViewSet(SingleArtifactContentUploadViewSet):
...
-                "action": ["list", "retrieve", "create"],
+                "action": ["list", "retrieve"],
+                "principal": "authenticated",
+                "effect": "allow",
+            },
+            {
+                "action": ["create"],
                 "principal": "authenticated",
                 "effect": "allow",
+                "condition": "has_required_repo_perms_on_upload:file.modify_filerepository",
```
ipanova, line 88: this is a nitpick, but if we follow the pattern we have established previously, once need to have modify *and* view repo perms. Same as - if you want to update a repo you need to have view and update perms

ipanova (approved):

PR #732, opened 2022-06-06
Port test crud content unit to pytest

bmbouter:
[noissue]


dkliban (commented):
The original tests include a test that ensures filtering a list of file content by relative path works. Please add taht to the CRUD test.

dkliban (approved):

PR #745, opened 2022-06-21
Remove the plugin removal test

goosemania:
This test requires a proper uninstallation of the plugin (due to a CI script which
checks for unapplied migrations) and then installing the plugin back from the commit
in the PR. The latter is not straightforward and is considered not worth the time and
effort, because the plugin removal is not a common operation and in most use cases
needed for the pulp-2to3-migration plugin only.

[noissue]


fao89:
I think we can easily modify this behavior on the single container

goosemania:
@fao89, I was told it was not worth the time/effort for what we get out of this testing. I'm fine both ways.
cc @bmbouter @dkliban @dralley

PR #749, opened 2022-06-22
Bump minor version

pulpbot:
[noissue]


fao89 (approved):

bmbouter:
We're actually evaluating just releasing the 1.10.next as forward compatible so I'm going to close this for now while we evaluate so it doesn't accidentally merge.

PR #754, opened 2022-06-29
Converts download policy tests to use pytest fixtures

dkliban:
[noissue]

### DIFF
--- a/.github/workflows/scripts/script.sh
+++ b/.github/workflows/scripts/script.sh
@@ -131,6 +131,7 @@ else
 fi
 export PULP_FIXTURES_URL="http://pulp-fixtures:8080"
 pushd ../pulp-cli
+pip install -r test_requirements.txt
 pytest -v -m pulp_file
 popd
 
--- a/pulp_file/tests/functional/api/test_download_policies.py
+++ b/pulp_file/tests/functional/api/test_download_policies.py
@@ -1,245 +1,76 @@
 """Tests for Pulp`s download policies."""
-from random import choice
-import unittest
+import pytest
 
-from pulp_smash.pulp3.bindings import delete_orphans, monitor_task
-from pulp_smash.pulp3.constants import ON_DEMAND_DOWNLOAD_POLICIES
+from pulp_smash.pulp3.bindings import monitor_task
 from pulp_smash.pulp3.utils import (
-    gen_repo,
     get_added_content_summary,
     get_content_summary,
 )
 
-from pulp_file.tests.functional.constants import (
-    FILE_FIXTURE_COUNT,
-    FILE_FIXTURE_SUMMARY,
-)
-from pulp_file.tests.functional.utils import (
-    gen_file_client,
-    gen_file_remote,
-    gen_pulpcore_client,
-    skip_if,
-)
+from pulp_file.tests.functional.utils import get_files_in_manifest
 
-from pulpcore.client.pulpcore import ArtifactsApi
-from pulpcore.client.pulp_file import (
-    ContentFilesApi,
-    FileFilePublication,
-    PublicationsFileApi,
-    RepositoriesFileApi,
-    RepositorySyncURL,
-    RemotesFileApi,
-)
+from pulpcore.client.pulp_file import FileFilePublication, RepositorySyncURL
 
 
-class SyncPublishDownloadPolicyTestCase(unittest.TestCase):
-    """Sync/Publish a repository with different download policies.
+@pytest.mark.parallel
+@pytest.mark.parametrize("download_policy", ["on_demand", "streamed"])
+def test_download_policy(
+    artifacts_api_client,
+    file_repo,
+    file_fixture_gen_remote_ssl,
+    file_remote_api_client,
+    file_repo_api_client,
+    file_pub_api_client,
+    basic_manifest_path,
+    gen_object_with_cleanup,
+    file_content_api_client,
+    download_policy,
+):
+    """Test that "on_demand" and "streamed" download policies work as expected."""
+    remote = file_fixture_gen_remote_ssl(manifest_path=basic_manifest_path, policy=download_policy)
+    file_repo = file_repo_api_client.read(file_repo.pulp_href)
+    assert file_repo.latest_version_href.endswith("/versions/0/")
 
-    This test targets the following issues:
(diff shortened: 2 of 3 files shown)


pulp_file/tests/functional/api/test_download_policies.py
```diff
@@ -2,244 +2,147 @@
...
-    RepositoriesFileApi,
-    RepositorySyncURL,
-    RemotesFileApi,
-)
-
+from pulp_file.tests.functional.utils import get_files_in_manifest
 
-class SyncPublishDownloadPolicyTestCase(unittest.TestCase):
-    """Sync/Publish a repository with different download policies.
+from pulpcore.client.pulp_file import FileFilePublication, RepositorySyncURL
```
bmbouter, line 13: Is `FileFilePublication` an object that just holds data like `RepositorySyncURL`? Aka, it's right that we don't need to have this provided by a pytest fixture and instead imported directly.

pulp_file/tests/functional/api/test_download_policies.py
```diff
@@ -2,244 +2,147 @@
...
+    file_fixture_gen_remote_ssl,
+    file_remote_api_client,
+    file_repo_api_client,
+    file_pub_api_client,
+    basic_manifest_path,
+    gen_object_with_cleanup,
+    file_content_api_client,
+):
     """
+    1. Create a File Remote with "on_demand" download policy.
```
bmbouter, line 28: I tend to put these in the code steps itself. What I worry about by having it up here is that it'll get out of date by not being close to the code.

pulp_file/tests/functional/api/test_download_policies.py
dkliban: That's correct.

pulp_file/tests/functional/api/test_download_policies.py
bmbouter: Oh I see those comments are more or less there. I'd say remove the docstring then (or rather convert it to a single line) as I think flake8 will require.

pulp_file/tests/functional/api/test_download_policies.py
```diff
@@ -2,244 +2,128 @@
...
-
-        # Assert that missing artifacts are downloaded
-        artifacts = artifacts_api.list().to_dict()["results"]
-        self.assertEqual(len(artifacts), FILE_FIXTURE_COUNT, artifacts)
+from pulp_file.tests.functional.utils import get_files_in_manifest
+
+from pulpcore.client.pulp_file import FileFilePublication, RepositorySyncURL
+
+
+def test_on_demand(
```
bmbouter, line 16: Can these be parallelized? If so mark them with `@pytest.mark.parallel`.

pulp_file/tests/functional/utils.py
```diff
@@ -292,3 +292,15 @@ def generate_manifest(name, file_list):
...
     return name
+
+
+def get_files_in_manifest(url):
+    """
+    Download a File Repository manifest and return content as a list of tuples.
+    [(name,sha256,size),]
+    """
+    files = []
+    with requests.get(url, verify=False) as r:
```
bmbouter, line 303: I believe our goal is to drop requests. Can this be done using `aiohttp` please?

pulp_file/tests/functional/utils.py
```diff
@@ -292,3 +294,21 @@ def generate_manifest(name, file_list):
...
+    files = []
+    r = asyncio.run(_download_manifest(url))
+    for line in r.splitlines():
+        files.append(tuple(line.decode().split(",")))
+    return files
+
+
+async def _download_manifest(url):
+    async with aiohttp.ClientSession() as session:
+        async with session.get(url, verify_ssl=False) as response:
```
bmbouter, line 313: Is the `verify_ssl=False` needed?

pulp_file/tests/functional/api/test_download_policies.py
```diff
@@ -1,245 +1,73 @@
...
+):
+    """Test that "on_demand" and "streamed" download policies work as expected."""
+    remote = file_fixture_gen_remote_ssl(manifest_path=basic_manifest_path, policy=download_policy)
+    file_repo = file_repo_api_client.read(file_repo.pulp_href)
+    assert file_repo.latest_version_href.endswith("/versions/0/")
+
+    # Check what content and artifacts are in the fixture repository
+    expected_files = get_files_in_manifest(remote.url)
+
+    body = RepositorySyncURL(remote=remote.pulp_href)
```
bmbouter, line 37: This little block could use a comment above it

pulp_file/tests/functional/api/test_download_policies.py
```diff
@@ -1,245 +1,73 @@
...
+    publish_data = FileFilePublication(repository=file_repo.pulp_href)
+    publication = gen_object_with_cleanup(file_pub_api_client, publish_data)
+    assert publication.repository_version is not None
+
+    # Change download policy to immediate and assert that artifacts are downloaded
+    response = file_remote_api_client.partial_update(remote.pulp_href, {"policy": "immediate"})
+    monitor_task(response.task)
+    remote = file_remote_api_client.read(remote.pulp_href)
+    assert remote.policy == "immediate"
+    monitor_task(file_repo_api_client.sync(file_repo.pulp_href, body).task)
```
bmbouter, line 71: This could probably become it's own little block with a comment.

pulp_file/tests/functional/utils.py
dkliban: Yes it is. aiohttp is not configured to trust the fixture cert.

.github/workflows/scripts/script.sh
```diff
@@ -131,6 +131,7 @@ else
 fi
 export PULP_FIXTURES_URL="http://pulp-fixtures:8080"
 pushd ../pulp-cli
+pip install -r test_requirements.txt
```
mdellweg, line 134: :+1:

pulp_file/tests/functional/api/test_download_policies.py
mdellweg: Can we switch of flake-docstring altogether? We are not using it on pulpcore, and it is not helping me write better code.

PR #756, opened 2022-07-01
Do not delete remotes

lubosmj:
Required PR: https://github.com/pulp/pulpcore/pull/2836
[noissue]

### DIFF
--- a/pulp_file/tests/functional/api/from_pulpcore/test_content_delivery.py
+++ b/pulp_file/tests/functional/api/from_pulpcore/test_content_delivery.py
@@ -10,6 +10,7 @@ from pulp_smash.pulp3.utils import (
 )
 
 from pulpcore.client.pulp_file import (
+    ApiException,
     RepositorySyncURL,
 )
 
@@ -30,10 +31,10 @@ def test_delete_remote_on_demand(
     gen_object_with_cleanup,
 ):
     # Create a remote with on_demand download policy
-    remote = file_fixture_gen_remote_ssl(manifest_path=basic_manifest_path, policy="on_demand")
+    remote1 = file_fixture_gen_remote_ssl(manifest_path=basic_manifest_path, policy="on_demand")
 
     # Sync from the remote
-    body = RepositorySyncURL(remote=remote.pulp_href)
+    body = RepositorySyncURL(remote=remote1.pulp_href)
     monitor_task(file_repo_api_client.sync(file_repo_with_auto_publish.pulp_href, body).task)
     repo = file_repo_api_client.read(file_repo_with_auto_publish.pulp_href)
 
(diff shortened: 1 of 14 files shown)


mdellweg:
If i read this correctly, it is accompanying a highly debatable implementation in pulpcore. In that case please mark this one as draft for now.

lubosmj:
The test is failing on the script `tests/scripts/pulp_file/test_publication.sh::publication.sh`. The functionality should not be affected and thus I am undrafting this PR.

pulp_file/tests/functional/api/from_pulpcore/test_content_delivery.py
```diff
@@ -43,20 +44,21 @@ def test_delete_remote_on_demand(
...
-    # Delete the remote and assert that downloading content returns a 404
-    monitor_task(file_remote_api_client.delete(remote.pulp_href).task)
-    with pytest.raises(ClientResponseError) as exc:
-        url = urljoin(distribution.base_url, expected_file_list[0][0])
-        download_file(url)
-    assert exc.value.status == 404
+    with pytest.raises(ApiException) as e:
+        # Delete the remote currently used by a repository version
+        monitor_task(file_remote_api_client.delete(remote1.pulp_href).task)
+        assert e.value.status == 406
```
bmbouter, line 52: I expected this to be outside of the context manager that way pytest.raises() is fully asserted first.

bmbouter:
Right now the pulp_file has only orphans leftover, but this would leave behind both orphans and remotes. IIRC, we agreed that was ok to change, but since then, I'm not so sure. I've gotten one request to have our test suite run against a production system, which we couldn't reasonably leave behind a lot of remotes on.

Another issue here I think I don't think every remote here is UUID generated to be recreatable every time, so if we start leaving them behind then we can't run the test suit N times in a row without a database reset. I spent a lot of time making that the case today so we can't regress. If we want to leave remotes behind we'd have to have each of them generated with random data.

So taken together, I think we have to clean up these remotes. One idea would be to have a cleanup at the end of the session type of fixture that creates remotes and then waits to delete them until the very end.

lubosmj:
> One idea would be to have a cleanup at the end of the session type of fixture that creates remotes and then waits to delete them until the very end.

This is probably the best idea. This can be also done at the end of every single test case if we do not want to write some mysterious hooks.

I will look into this fixture: https://github.com/pulp/pulp-smash/blob/ca6c35d1455f39e5e8a618bc7833624f9d7d5983/pulp_smash/pulp3/pytest_plugin/__init__.py#L469 to see whether it is possible to clean up remotes at the end, after removing all objects in the list.

bmbouter:
> > One idea would be to have a cleanup at the end of the session type of fixture that creates remotes and then waits to delete them until the very end.
> 
> This is probably the best idea. This can be also done at the end of every single test case if we do not want to write some mysterious hooks.
> 
> I will look into this fixture: https://github.com/pulp/pulp-smash/blob/ca6c35d1455f39e5e8a618bc7833624f9d7d5983/pulp_smash/pulp3/pytest_plugin/__init__.py#L469 to see whether it is possible to clean up remotes at the end, after removing all objects in the list.

The issue is that fixture is not session level scoped, and I don't think you can switch it because we need it to be function scoped as it is today. So you need a new fixture that performs create and cleanup that is session level scoped. @dkliban @mdellweg are you thinking similarly?

@lubosmj is this making sense?

PR #757, opened 2022-07-02
Refactors the auto publish and distribute tests to pytest fixtures

dkliban:
[noissue]

### DIFF
--- a/pulp_file/tests/functional/api/test_auto_publish.py
+++ b/pulp_file/tests/functional/api/test_auto_publish.py
@@ -1,11 +1,10 @@
 """Tests that sync file plugin repositories."""
-import unittest
+import pytest
 
-from pulp_smash import config
 from pulp_smash.pulp3.bindings import delete_orphans, monitor_task
-from pulp_smash.pulp3.utils import delete_version, download_content_unit, gen_repo
+from pulp_smash.pulp3.utils import gen_repo
 
-from pulp_file.tests.functional.utils import gen_file_client, gen_file_remote
+from pulp_file.tests.functional.utils import gen_file_client, get_files_in_manifest
 
 from pulpcore.client.pulp_file import (
     ContentFilesApi,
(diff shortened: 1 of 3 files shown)


pulp_file/tests/functional/conftest.py
```diff
@@ -92,6 +92,13 @@ def file_repo(file_repo_api_client, gen_object_with_cleanup):
     return gen_object_with_cleanup(file_repo_api_client, gen_repo())
 
 
+@pytest.fixture
+def file_repo_with_auto_publish(file_repo_api_client, gen_object_with_cleanup):
+    return gen_object_with_cleanup(
+        file_repo_api_client, gen_repo(autopublish=True, manifest="TEST_MANIFEST")
+    )
+
+
```
mdellweg, line 101: I'd say, this fixture is quite special to the autopublish test. We should move it there.

pulp_file/tests/functional/api/test_auto_publish.py
```diff
@@ -17,94 +15,71 @@
...
+    repo = file_repo_api_client.read(repo.pulp_href)
+
+    # Assert that a new repository version was created and a publication was created
+    assert repo.latest_version_href.endswith("/versions/1/")
+    assert file_pub_api_client.list(repository=repo.pulp_href).count == 1
+    assert file_pub_api_client.list(repository_version=repo.latest_version_href).count == 1
+
+    # Assert that the publication has a custom manifest
+    publication = file_pub_api_client.list(repository_version=repo.latest_version_href).results[0]
+    assert publication.manifest == "TEST_MANIFEST"
```
mdellweg, line 60: It's hard to understand where this comes from because the fixture generating this is far away.

pulp_file/tests/functional/api/test_auto_publish.py
```diff
@@ -17,94 +15,71 @@
...
+    files_in_second_publication = get_files_in_manifest(
+        "{}{}".format(distribution.base_url, publication.manifest)
+    )
+    assert repo.latest_version_href.endswith("/versions/2/")
+    assert file_pub_api_client.list(repository=repo.pulp_href).count == 2
+    assert file_pub_api_client.list(repository_version=repo.latest_version_href).count == 1
+    assert len(files_in_second_publication - files_in_first_publication) == 1
+    assert (files_in_second_publication - files_in_first_publication).pop()[
+        1
+    ] == file_random_content_unit.sha256
```
mdellweg, line 85: Maybe this formats better if you saved the difference set in a variable.

mdellweg (commented):
Small asks for style.
But great rewrite!

pulp_file/tests/functional/api/test_auto_publish.py
dkliban: I moved the fixture to this file.

pulp_file/tests/functional/api/test_auto_publish.py
```diff
@@ -17,94 +16,77 @@
...
+    repo = file_repo_api_client.read(repo.pulp_href)
+    files_in_second_publication = get_files_in_manifest(
+        "{}{}".format(distribution.base_url, publication.manifest)
+    )
+    files_added = files_in_second_publication - files_in_first_publication
+    assert repo.latest_version_href.endswith("/versions/2/")
+    assert file_pub_api_client.list(repository=repo.pulp_href).count == 2
+    assert file_pub_api_client.list(repository_version=repo.latest_version_href).count == 1
+    assert len(files_added) == 1
+    assert files_added.pop()[1] == file_random_content_unit.sha256
```
mdellweg, line 92: Sorry for nitpicking again. `pop` changes the `files_added` and i know that is not a problem programmatically, but semantically i shouldn't. Can you try this (?):
```suggestion
    assert list(files_added)[0][1] == file_random_content_unit.sha256
```

mdellweg (approved):
Left a suggestion, but fine to merge anyway.

pulp_file/tests/functional/api/test_auto_publish.py
dkliban: I added this change.

PR #761, opened 2022-07-06
Refactors ACS tests to use pytest fixtures

dkliban:
[noissue]

### DIFF
--- a/pulp_file/tests/functional/api/test_acs.py
+++ b/pulp_file/tests/functional/api/test_acs.py
@@ -1,166 +1,227 @@
 import hashlib
-import unittest
-from random import choice
+import pytest
+import uuid
 from urllib.parse import urljoin
 
-from pulp_smash import config, utils
 from pulp_smash.pulp3.bindings import (
     PulpTaskError,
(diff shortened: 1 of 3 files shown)


pulp_file/tests/functional/api/test_acs.py
```diff
@@ -1,166 +1,237 @@
...
-        pulp_hash = hashlib.sha256(content).hexdigest()
-
-        self.assertEqual(fixtures_hash, pulp_hash)
+@pytest.mark.parallel
+def test_acs_path_validation(
+    file_acs_api_client, file_fixture_gen_remote, basic_manifest_path, gen_object_with_cleanup
+):
+    """Test that paths starting with "/" are not accepted by ACS API."""
+    remote = file_fixture_gen_remote(manifest_path=basic_manifest_path, policy="on_demand")
+    acs_data = {"name": "foo", "remote": remote.pulp_href, "paths": ["good/path", "/bad/path"]}
```
bmbouter, line 28: This test probably only needs `/bad/path`.

Issue #630, opened 2021-12-20
Commas in the relative_path field of the PULP_MANIFEST cause a validation error when that manifest is used as the target of a pulp_file remote


pulpbot:
Author: ajsween (ajsween)

Redmine Issue: 9207, https://pulp.plan.io/issues/9207

---

There are no errors thrown when adding file content to a repository where the file name/relative path contains a comma. The PULP_MANIFEST is then updated to include a comma in the relative_path field thus breaking deserialization of the PULP_MANIFEST when referenced as the file remote target (URL) of an external pulp repository.

Example:
~~~ text
All/bison-3.7.6,1.txz,6c7d3a44ef1f8dc77a1ff8b6df38c4a41b4b443e6ea433e662805a5e006268ec,489308
~~~

Commas are valid filename characters in both Linux and Windows filesystems. FreeBSD specifically uses a versioning schema that utilizes commas in the filenames of packages to denote [PORTEPOCH](https://docs.freebsd.org/en/books/porters-handbook/makefiles/#:~:text=5.2.3.2.-,PORTEPOCH,-From%20time%20to)

The fix should be either to serialize the fields using double quotes in accordance with [RFC4180](https://datatracker.ietf.org/doc/html/rfc4180#:~:text=6.%20%20Fields%20containing%20line%20breaks%20(CRLF)%2C%20double%20quotes%2C%20and%20commas%0A%20%20%20%20%20%20%20should%20be%20enclosed%20in%20double-quotes.%20%20For%20example%3A%0A%0A%20%20%20%20%20%20%20%22aaa%22%2C%22b%20CRLF%0A%20%20%20%20%20%20%20bb%22%2C%22ccc%22%20CRLF%0A%20%20%20%20%20%20%20zzz%2Cyyy%2Cxxx), URL encode commas in the field, or choose a different serialization format (e.g. YAML, JSON, XML, etc...).

I tested using URL encoding of the commas ```%2C``` which allowed the PULP_MANIFEST to successfully be validated and files to be synced but then prevented files from being downloaded from the sync'd repository due to how the file distribution serves file artifacts.

pulpbot:
From: ajsween (ajsween)
Date: 2021-08-04T03:12:43Z

---

[...]

PR #767, opened 2022-07-11
The relative_path field of PULP_MANIFEST can now contain commas

MichalPysik:
Commas are a valid filename characters in both Linux and Windows
file systems, but the previous way of generating manifest file
did not	allow them. This commit	fixes this issue, and introduces
a test for this	very specific edge-case. The test is split into
all three possible sub-cases: comma in filename, comma in folder
name, comma(s) in both.

closes pulp#630

### DIFF
--- a/pulp_file/manifest.py
+++ b/pulp_file/manifest.py
@@ -2,6 +2,8 @@ from collections import namedtuple
 
 from gettext import gettext as _
 
+from re import fullmatch
+
 
 Line = namedtuple("Line", ("number", "content"))
 
@@ -49,8 +51,15 @@ class Entry:
             ValueError: on parsing error.
 
         """
-        part = [s.strip() for s in line.content.split(",")]
-        if len(part) != 3:
+        all_parts = line.content.count(",") >= 2
+        if all_parts:
+            relative_path, digest, size = [s.strip() for s in line.content.rsplit(",", maxsplit=2)]
+        if (
+            not all_parts
+            or not fullmatch(r"^[^/]+(/[^/]+)*$", relative_path)
+            or not fullmatch(r"^[0-9a-fA-F]+$", digest)
+            or not size.isdigit()
+        ):
             raise ValueError(
                 _(
                     "Error: Parsing of the manifest file failed on line:{n}.\n"
@@ -59,7 +68,7 @@ class Entry:
                     "composed of lines in the following format: <relative_path>,<digest>,<size>."
                 ).format(n=line.number)
             )
-        return Entry(relative_path=part[0], digest=part[1], size=int(part[2]))
+        return Entry(relative_path=relative_path, digest=digest, size=int(size))
 
     def __str__(self):
         """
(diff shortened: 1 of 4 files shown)


lubosmj:
@MichalPysik, to get rid of the error, please, replace "closes pulp#630" with "closes #630" in the commit message.

MichalPysik:
> @MichalPysik, to get rid of the error, please, replace `closes pulp#630` with `closes #630` in the commit message.

@lubosmj thanks a lot!

MichalPysik:
> 
That was my initial solution, but we discussed further and decided that since each part should also have a separate check (filepath is relative and refers to a file, not folder, therefore cannot start or end with '/', I'm also not sure which characters can and cannot be part of a digest), it might be more elegant to perform the check in a single line.

dralley:
On second thought I doubt we actually want that kind of granular feedback.  I still have mixed feelings about the regex but I can live with it if we feel like it's a more complete approach.  If we do go with it, let's make sure it's well documented what each portion is doing so that breaking out a regex reference isn't necessary :)

 I'll let others weigh in.

bmbouter:
I'm not sure what we loose by not using the regex, but I'm hesitant to use the regex.

MichalPysik:
@bmbouter @dralley The regex made sense when I wasn't sure whether the digest can also contain commas, but since now I know it should be in the form of a hexadecimal number, I will rewrite the code tommorow using rsplit() and an individual check for each part, which should also make it more readable.

pulp_file/manifest.py
```diff
@@ -49,8 +51,12 @@ def parse(line):
             ValueError: on parsing error.
 
         """
-        part = [s.strip() for s in line.content.split(",")]
-        if len(part) != 3:
+        relative_path, digest, size = [s.strip() for s in line.content.rsplit(",", maxsplit=2)]
```
mdellweg, line 54: The subtle problem with this solution is, if the line does not contain enough commas, it will fail here and not raise the ValueError.

BTW: This would have been covered nicely by the original full-line regex.

pulp_file/manifest.py
```diff
@@ -49,8 +51,12 @@ def parse(line):
...
+        if (
+            not fullmatch(r"^[^/]+(/[^/]+)*$", relative_path)
```
mdellweg, line 56: Can you explain, why you use this regex? IMHO it's still too complicated.

pulp_file/manifest.py
MichalPysik: Thanks for the heads up, I didn't notice that, I will modify that.

pulp_file/manifest.py
MichalPysik: Because apart from checking that both the first and the last character is different from '/', I also wanted to check that there is no "//", for example "aaa///ff" should be invalid.

mdellweg: ahh, i see.

mdellweg (approved):

dralley (approved):

PR #769, opened 2022-07-12
Refactors content delivery tests to use pytest fixtures

dkliban:
[noissue]

### DIFF
--- a/pulp_file/tests/functional/conftest.py
+++ b/pulp_file/tests/functional/conftest.py
@@ -1,5 +1,6 @@
 import aiofiles
 import logging
+import os
 import uuid
 
 from aiohttp import web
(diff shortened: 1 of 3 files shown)


pulp_file/tests/functional/conftest.py
```diff
@@ -125,6 +126,23 @@ def basic_manifest_path(write_3_iso_file_fixture_data_factory):
     return write_3_iso_file_fixture_data_factory("basic")
 
 
+@pytest.fixture
+def copy_manifest_only(file_fixtures_root):
```
bmbouter, line 130: By convention, I am encouraging everyone who uses the factory pattern to end their fixture name with `_factory`.

bmbouter (approved):
This looks like it's testing all the right things!

PR #770, opened 2022-07-13
Converts rbac contet guard tests to pytest fixtures

dkliban:
The test logic is now slightly changed. The creator user is given the "file.filedistribution_creator"
so that they can create and modify the distribution also.

Required PR: https://github.com/pulp/pulpcore/pull/2966

[noissue]


gerrod3 (approved):
This looks good, I wonder if the changes to the get_url helper will affect other tests. I remember specifically for requests I had to set it to not read the .netrc file and I'm not sure if aiohttp is the same.

dkliban:
@gerrod3 `trust_env=False` by default. https://docs.aiohttp.org/en/stable/client_reference.html#aiohttp.ClientSession

pulp_file/tests/functional/api/from_pulpcore/test_content_guard.py
```diff
@@ -1,139 +1,88 @@
...
+    creator_user = gen_user(
+        model_roles=["core.rbaccontentguard_creator", "file.filedistribution_creator"]
+    )
+    user_a = gen_user()
+    user_b = gen_user()
+
+    all_users = [creator_user, user_a, user_b, admin_user, None]
+    group = gen_object_with_cleanup(groups_api_client, {"name": str(uuid.uuid4())})
+    gen_object_with_cleanup(groups_users_api_client, group.pulp_href, {"username": user_b.username})
+    gen_object_with_cleanup(groups_users_api_client, group.pulp_href, {"username": user_a.username})
```
mdellweg, line 37: We autodelete the group and the users. No need for `gen_object_with_cleanup`.

pulp_file/tests/functional/api/from_pulpcore/test_content_guard.py
```diff
@@ -1,139 +1,88 @@
...
+    gen_object_with_cleanup,
+):
+    # Create all of the users and groups
+    creator_user = gen_user(
+        model_roles=["core.rbaccontentguard_creator", "file.filedistribution_creator"]
+    )
+    user_a = gen_user()
+    user_b = gen_user()
+
+    all_users = [creator_user, user_a, user_b, admin_user, None]
```
mdellweg, line 34: Can this be `anonymous_user` instead of `None`?

dkliban: yes, let's use anonymous_user.

pulp_file/tests/functional/api/from_pulpcore/test_content_guard.py
dkliban: Sounds good. I'll change it.

PR #773, opened 2022-07-14
Convert CRUD Repo tests to pytest

gerrod3:
[noissue]

### DIFF
--- a/pulp_file/tests/functional/api/from_pulpcore/test_crud_repos.py
+++ b/pulp_file/tests/functional/api/from_pulpcore/test_crud_repos.py
@@ -1,216 +1,136 @@
 """Tests that CRUD repositories."""
 import json
+import pytest
 import re
-import time
-import unittest
+from aiohttp import BasicAuth
 from itertools import permutations
+from subprocess import run
 from urllib.parse import urljoin
 
-from pulp_smash import api, cli, config, utils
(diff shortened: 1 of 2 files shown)


pulp_file/tests/functional/api/from_pulpcore/test_crud_repos.py
```diff
@@ -225,234 +145,185 @@ def setUp(self):
...
-        }
-        self.raise_for_invalid_request(remote_attrs)
+        with pytest.raises(ApiException) as ae:
+            file_remote_api_client.create(remote_attrs)
+            assert ae.value.status == 400
+
+    # Test the validation of an invalid absolute pathname.
+    remote_attrs = {
+        "name": utils.uuid4(),
+        "url": "file://error/path/name",
```
dkliban, line 300: I know that this is just a port of the existing test, but I don't understand the difference between this and the previous assertion.

dkliban (approved):
I left one comment.

pulp_file/tests/functional/api/from_pulpcore/test_crud_repos.py
mdellweg: There is one / missing. But maybe we should use
```suggestion
        "url": "file://tmp/good",
```
to only have one reason to fail.

PR #780, opened 2022-07-25
Extends tests for Publications

dkliban:
This patch removes the from_pulpcore.test_crd_publications module. The test_publications module
has been augmented with tests that were in from_pulpcore.test_crd_publications module.

[noissue]


pulp_file/tests/functional/api/test_publish.py
```diff
@@ -61,3 +67,68 @@ def test_creating_publications(
...
+        "pulp_href",
+        "pulp_created",
+        "repository",
+        "repository_version",
+        "distributions",
+    )
+    config = file_repo_api_client.api_client.configuration
+    auth = BasicAuth(login=config.username, password=config.password)
+    full_href = urljoin(config.host, publication.pulp_href)
+    for field_pair in permutations(fields, 2):
```
mdellweg, line 86: Will this iterate over all ordered pairs? (20 combinations)

dkliban: Yes, the original test only did 6 combinations. It selected 2 fields from "pulp_created", "pulp_href", and "distributions". I don't feel too strongly about it. Should I change it back to these 3 fields?

mdellweg: What are we testing here? I think it's the `fields=` querystring feature. Why not just select 3 combinations and test them explicitly?

dkliban: I am ok with that. We should probably do the same thing for the other CRUD tests also.

mdellweg (approved):

PR #789, opened 2022-08-04
Allow uploading existing FileContent units

### LINKED ISSUE #774: Reuse existing artifact and file content units in case the user provided proof that they have the bits
On content upload, pulp should not reject the request for any uniqueness constraint violation, but realize that the user was able to provide the bits and therefor grant using the existing records.

```
# Upload first time works
pulp file content upload --relative-path "README.md" --file "README.md" --chunk-size 100
# Uploading a second time fails
pulp file content upload --relative-path "README.md" --file "README.md" --chunk-size 100
# Uploading for a different content_unit works again
pulp file content upload --relative-path "README2.md" --file "README.md" --chunk-size 100
```

edit: All of the above commands should succeed.

`--chunk-size` is there for a different issue in pulp-cli.

MichalPysik:
Attemting to upload a FileContent unit already present inside a
different repository no longer results in an exception being raised,
the existing unit is re-used instead (without notifying the user).

An exception is still raised if either a user attempts to upload the
unit into a repository inside which it already exists, or the admin
attempts to upload an existing unit without specifying a repository.

closes #774

### DIFF
--- a/CHANGES/774.feature
+++ b/CHANGES/774.feature
@@ -1,0 +1,1 @@
+The upload feature was changed to accept already existing file content. This allows multiple users to own identical files.
--- a/pulp_file/app/serializers.py
+++ b/pulp_file/app/serializers.py
@@ -35,21 +35,13 @@ class FileContentSerializer(SingleArtifactContentUploadSerializer, ContentChecks
 
         data["digest"] = data["artifact"].sha256
 
-        content = FileContent.objects.filter(
-            digest=data["digest"], relative_path=data["relative_path"]
-        )
-
-        if content.exists():
-            content.get().touch()
-
-            raise serializers.ValidationError(
-                _(
-                    "There is already a file content with relative path '{path}' and digest "
-                    "'{digest}'."
-                ).format(path=data["relative_path"], digest=data["digest"])
-            )
-
         return data
+
+    def retrieve(self, validated_data):
+        content = FileContent.objects.filter(
+            digest=validated_data["digest"], relative_path=validated_data["relative_path"]
+        )
+        return content.first()
 
     class Meta:
         fields = (
--- a/pulp_file/tests/functional/api/test_crud_content_unit.py
+++ b/pulp_file/tests/functional/api/test_crud_content_unit.py
@@ -8,6 +8,9 @@ from pulp_smash.pulp3.bindings import (
     monitor_task,
     PulpTaskError,
 )
+from pulp_smash.pulp3.utils import get_content_summary, get_added_content_summary
+
+from pulpcore.client.pulpcore.exceptions import ApiException as coreApiException
 
 
 @pytest.mark.parallel
@@ -38,14 +41,62 @@ def test_crud_content_unit(
 
 
 @pytest.mark.parallel
-def test_same_sha256_same_relative_path_not_allowed(
+def test_same_sha256_same_relative_path_no_repo(
     random_artifact, file_content_api_client, gen_object_with_cleanup
 ):
     artifact_attrs = {"artifact": random_artifact.pulp_href, "relative_path": str(uuid.uuid4())}
-    gen_object_with_cleanup(file_content_api_client, **artifact_attrs)
 
-    with pytest.raises(PulpTaskError):
-        gen_object_with_cleanup(file_content_api_client, **artifact_attrs)
+    content1 = file_content_api_client.read(
+        monitor_task(file_content_api_client.create(**artifact_attrs).task).created_resources[0]
+    )
+    content2 = file_content_api_client.read(
+        monitor_task(file_content_api_client.create(**artifact_attrs).task).created_resources[0]
+    )
+    assert content1.pulp_href == content2.pulp_href
+    assert file_content_api_client.read(content1.pulp_href).pulp_href == content2.pulp_href
+
+
+@pytest.mark.parallel
+def test_same_sha256_same_relative_path_repo_specified(
+    random_artifact,
+    file_content_api_client,
+    file_repo_api_client,
+    gen_user,
+    file_fixture_gen_file_repo,
+):
+    max = gen_user(model_roles=["file.filerepository_creator"])
+    john = gen_user(model_roles=["file.filerepository_creator"])
+
+    with max:
+        repo1 = file_fixture_gen_file_repo(name=str(uuid.uuid4()))
+    with john:
+        repo2 = file_fixture_gen_file_repo(name=str(uuid.uuid4()))
+
+    artifact_attrs = {"artifact": random_artifact.pulp_href, "relative_path": str(uuid.uuid4())}
+
+    artifact_attrs["repository"] = repo1.pulp_href
+    with max:
+        response1 = file_content_api_client.create(**artifact_attrs)
+        response2 = file_content_api_client.create(**artifact_attrs)
+
+    content1 = file_content_api_client.read(monitor_task(response1.task).created_resources[1])
+    content2 = file_content_api_client.read(monitor_task(response2.task).created_resources[0])
+    assert content1.pulp_href == content2.pulp_href
+    repo1 = file_repo_api_client.read(repo1.pulp_href)
+    assert repo1.latest_version_href.endswith("/versions/1/")
+    assert get_content_summary(repo1.to_dict()) == {"file.file": 1}
+    assert get_added_content_summary(repo1.to_dict()) == {"file.file": 1}
+
+    artifact_attrs["repository"] = repo2.pulp_href
+    with john:
+        ctask3 = file_content_api_client.create(**artifact_attrs).task
+
+    content3 = file_content_api_client.read(monitor_task(ctask3).created_resources[1])
+    assert content3.pulp_href == content1.pulp_href
+    repo2 = file_repo_api_client.read(repo2.pulp_href)
+    assert repo2.latest_version_href.endswith("/versions/1/")
+    assert get_content_summary(repo2.to_dict()) == {"file.file": 1}
+    assert get_added_content_summary(repo2.to_dict()) == {"file.file": 1}
 
 
 @pytest.mark.parallel
@@ -151,9 +202,31 @@ def test_create_file_content_from_chunked_upload(
         uploads_api_client.update(
             upload_href=upload.pulp_href, file=file_2, content_range="bytes 128-255/256"
         )
+        most_recent_path = str(uuid.uuid4())
         response = file_content_api_client.create(
-            upload=upload.pulp_href, relative_path=str(uuid.uuid4())
+            upload=upload.pulp_href, relative_path=most_recent_path
         )
         task = monitor_task(response.task)
         content = file_content_api_client.read(task.created_resources[0])
         assert content.sha256 == expected_digest
+        # Upload gets deleted if the content gets created
+        with pytest.raises(coreApiException):
+            uploads_api_client.read(upload.pulp_href)
+
+    # Attempt to create a duplicate content by re-using the most recent relative path
+    upload = gen_object_with_cleanup(uploads_api_client, {"size": 256})
+    uploads_api_client.update(
+        upload_href=upload.pulp_href, file=file_1, content_range="bytes 0-127/256"
+    )
+    uploads_api_client.update(
+        upload_href=upload.pulp_href, file=file_2, content_range="bytes 128-255/256"
+    )
+    response = file_content_api_client.create(
+        upload=upload.pulp_href, relative_path=most_recent_path
+    )
+    task = monitor_task(response.task)
+    content = file_content_api_client.read(task.created_resources[0])
+    assert content.sha256 == expected_digest
+    # Upload gets deleted even though no new content got created
+    with pytest.raises(coreApiException):
+        uploads_api_client.read(upload.pulp_href)
--- a/pulp_file/tests/unit/test_serializers.py
+++ b/pulp_file/tests/unit/test_serializers.py
@@ -45,11 +45,11 @@ class TestFileContentSerializer(TestCase):
         self.assertFalse(serializer.is_valid())
 
     def test_duplicate_data(self):
-        """Test that the FileContentSerializer does not accept data."""
+        """Test that the FileContentSerializer accepts duplicate valid data."""
         FileContent.objects.create(relative_path="foo", digest=self.artifact.sha256)
         data = {
             "artifact": f"{V3_API_ROOT}artifacts/{self.artifact.pk}/",
             "relative_path": "foo",
         }
         serializer = FileContentSerializer(data=data)
-        self.assertFalse(serializer.is_valid())
+        self.assertTrue(serializer.is_valid())
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,1 +1,1 @@
-pulpcore>=3.21.0.dev,<3.25
+pulpcore>=3.22.0.dev,<3.25


MichalPysik:
Marked this as draft for know, since I'm currently working on new test cases relevant to this feature as Ina suggested.
I opened this PR early so you guys can check it out and give me some input in case you have a suggestion.

gerrod3 (commented):
Looking good. Some questions to discuss:

1. Should this check happen inside the task or before the task? I feel like this check is more appropriate before the create task is dispatched that way the feedback is more immediate.
2. Should the admin user get an exception for already existing content without a repository? Since admins can see all content I feel returning the existing content href would be more helpful.
3. If the content already exists and is already in the repository should we tell the user or just let the create task do no work? I think this is more subjective, but I'm in favor of informing the user since I think having a task that does no work is confusing.
4. Should this live here or in pulpcore? I feel that this can be made generic enough to be used by all plugins with single artifact content.

Sorry for all the questions, but I feel we (the RTFM group) didn't fully flesh out this idea before writing it down into an issue.

pulp_file/app/serializers.py
```diff
@@ -35,21 +35,52 @@ def deferred_validate(self, data):
...
+            content.touch()
+            if not repository:
+                raise serializers.ValidationError(
+                    _(
+                        "There is already a file content with relative path '{path}' and digest "
+                        "'{digest}'. No action has been done since no repository was specified."
+                    ).format(path=validated_data["relative_path"], digest=validated_data["digest"])
+                )
+        else:
+            content = super(FileContentSerializer.__bases__[0], self).create(validated_data)
```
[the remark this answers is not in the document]
MichalPysik: I wanted to call the bas parent's variant (which was already reimplemented by __bases__[1]), so I could not use super().

MichalPysik:
> Looking good. Some questions to discuss:
> 
>     1. Should this check happen inside the task or before the task? I feel like this check is more appropriate before the create task is dispatched that way the feedback is more immediate.
> 
>     2. Should the admin user get an exception for already existing content without a repository? Since admins can see all content I feel returning the existing content href would be more helpful.
> 
>     3. If the content already exists and is already in the repository should we tell the user or just let the create task do no work? I think this is more subjective, but I'm in favor of informing the user since I think having a task that does no work is confusing.
> 
>     4. Should this live here or in pulpcore? I feel that this can be made generic enough to be used by all plugins with single artifact content.
> 
> 
> Sorry for all the questions, but I feel we (the RTFM group) didn't fully flesh out this idea before writing it down into an issue.


1. The check cannot happen before during the serializer validation (as it was till now), since there is no context available, so we can only either allow or deny all duplicates there.

2. I think this is a good point, I'm probably going to change that.

3. This was exactly my though process, I believe throwing an exception is better than a task that does nothing. I believe returning the existing href the same way as in the admin case is pointless, since users don't care about hrefs  (but please correct me if I'm wrong) and checking if it was the admin who created the request is too much of an edge case to cover.

4. I brought up this question at the latest sprint planning and @mdellweg suggested to keep it inside pulp_file for now, but if different folks have different views on this, maybe let's go for a team-wide discussion about this topic.

ipanova:
> Looking good. Some questions to discuss:
> 
>     1. Should this check happen inside the task or before the task? I feel like this check is more appropriate before the create task is dispatched that way the feedback is more immediate.
> 
>     2. Should the admin user get an exception for already existing content without a repository? Since admins can see all content I feel returning the existing content href would be more helpful.

+1
> 
>     3. If the content already exists and is already in the repository should we tell the user or just let the create task do no work? I think this is more subjective, but I'm in favor of informing the user since I think having a task that does no work is confusing.

I think we should follow the same pattern as everywhere - if the user has proven that he has the bits to the content then we should whether re-use existing content or return the href for it.
> 
>     4. Should this live here or in pulpcore? I feel that this can be made generic enough to be used by all plugins with single artifact content.
> 
> 
> Sorry for all the questions, but I feel we (the RTFM group) didn't fully flesh out this idea before writing it down into an issue.

MichalPysik:
Discussed further with @ipanova, and the most appropriate solution was to not throw any errors at all, and just return the existing href for both specified/unspecified repository. The only expected exception now is when normal user does not specify repository, but that is an rbac error, not content duplicity error.

mdellweg:
> I think we should follow the same pattern as everywhere - if the user has proven that he has the bits to the content then we should whether re-use existing content or return the href for it.

I agree completely. We should try not to make any behaviour dependent on weather the request was made by the admin or not. That would be so confusing.

CHANGES/774.feature
```diff
@@ -0,0 +1 @@
+Existing FileContent units can be re-uploaded into a different repository (without the user knowing it already exists elsewhere). The tests have been updated to account for this feature.
```
mdellweg, line 1: This is meant for end-users and admins. They don't care about us writing tests. Also we write changelogs in past tense.
The "into a repository" part is not crucial to this change, and it does not need to be a different one for sure. I would say something like "The upload feature was changed to accept already existing file content. This allows users in a multi user setup to own identical files.".

pulp_file/app/serializers.py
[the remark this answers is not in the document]
mdellweg: Let's not do this. It's a sign of improper inheritance. Also we are stretching the notion of "create" quite far at this point. What this function is doing is "get-or-create-and-add-to-repository". I would rewrite it as `get_or_create`. And then we will see how we can move most of the generics into `pulpcore/plugin`.

pulp_file/tests/unit/test_serializers.py
```diff
@@ -35,21 +35,21 @@ def test_valid_data(self):
         serializer = FileContentSerializer(data=data)
         self.assertTrue(serializer.is_valid())
 
-    def test_absolute_path_data(self):
-        """Test that the FileContentSerializer does not accept data."""
```
mdellweg, line 39: This diff is confusing. Please try not to touch the test you are not changing.

MichalPysik: I did change the test - it was making sure that there can never be duplicates, but this new feature allows them, so I had to "invert" the test (and I switched its order in the file for consistency, that's why the diff seems a bit long.

pulp_file/app/serializers.py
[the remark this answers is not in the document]
MichalPysik: - well, I need to the original method from bases[0], which has been reimplemented in bases[1], and we are currently in bases[3] and need our own implementation without calling bases[1] (==bases[2]), so I don't know any other way
- the method is called by name and I don't have access to the called calling it, so I could probably only rename the method, but still have to create a method calling it called "create" like this:

def get_or_create():
    ...code...

def create():
    get_or_create()

CHANGES/774.feature
MichalPysik: Thanks, I didn't realize the changelogs should be more end-user-functionality focused and not mention invisible backend changes. Just updated that.

pulp_file/tests/unit/test_serializers.py
MichalPysik: @mdellweg I didn't realize it __bases__[0] actually pointed to the same class as super(), so I though I was calling a different method all along. Replaced it by super() now and everything should still work.

mdellweg: Thinking about git history, it would be better to keep the tests in their original order.

MichalPysik: Fixed.

pulp_file/tests/functional/api/test_crud_content_unit.py
```diff
@@ -38,14 +40,49 @@ def test_crud_content_unit(
...
-    random_artifact, file_content_api_client, gen_object_with_cleanup
+def test_same_sha256_same_relative_path_no_repo(
+    random_artifact, file_content_api_client, gen_object_with_cleanup, gen_user
 ):
     artifact_attrs = {"artifact": random_artifact.pulp_href, "relative_path": str(uuid.uuid4())}
-    gen_object_with_cleanup(file_content_api_client, **artifact_attrs)
 
-    with pytest.raises(PulpTaskError):
-        gen_object_with_cleanup(file_content_api_client, **artifact_attrs)
+    content1 = gen_object_with_cleanup(file_content_api_client, **artifact_attrs)
```
mdellweg, line 48: There's no use for gen_object_with_cleanup on content.

pulp_file/tests/functional/api/test_crud_content_unit.py
```diff
@@ -38,14 +40,49 @@ def test_crud_content_unit(
...
+    content2 = gen_object_with_cleanup(file_content_api_client, **artifact_attrs)
+    assert content1.pulp_href == content2.pulp_href
+
+    charlie = gen_user()
+    with charlie:
+        with pytest.raises(ApiException):
+            gen_object_with_cleanup(file_content_api_client, **artifact_attrs)
```
mdellweg, line 55: Why is this expected to fail? Please leave a comment. Also isn't this covered elsewhere already?

pulp_file/tests/functional/api/test_crud_content_unit.py
```diff
@@ -38,14 +40,49 @@ def test_crud_content_unit(
...
+    with john:
+        repo2 = file_fixture_gen_file_repo(name=str(uuid.uuid4()))
+
+    artifact_attrs = {"artifact": random_artifact.pulp_href, "relative_path": str(uuid.uuid4())}
+
+    artifact_attrs["repository"] = repo1.pulp_href
+    with max:
+        content1 = gen_object_with_cleanup(file_content_api_client, **artifact_attrs)
+        content2 = gen_object_with_cleanup(file_content_api_client, **artifact_attrs)
+    assert content1.pulp_href == content2.pulp_href
```
mdellweg, line 80: Can you add a check, that the repository got exactly on new version containing exactly that piece of content?

pulp_file/tests/functional/api/test_crud_content_unit.py
MichalPysik: Well only admin can upload content without specifying a repository. It tests permission and not duplicity, so while still relevant, it is a bit redundant here. I might remove it.

pulp_file/tests/functional/api/test_crud_content_unit.py
MichalPysik: I just followed the pattern found in other tests in the same file.

pulp_file/tests/functional/api/test_crud_content_unit.py
MichalPysik: This is something I actually attempted, but could not find a way to do. But I also believe it is quite an important thing, so I should probably try to figure out a way.

pulp_file/tests/functional/api/test_crud_content_unit.py
gerrod3: To be clear the reason why you shouldn't use `gen_object_with_cleanup` on content is because content doesn't have a delete endpoint. `gen_object_with_cleanup` is defined here: https://github.com/pulp/pulp-smash/blob/main/pulp_smash/pulp3/pytest_plugin/__init__.py#L468 and is to be used for any object that has both a create and delete endpoint.

pulp_file/tests/functional/api/test_crud_content_unit.py
gerrod3: Hint: the File Repository Version endpoint might have what you are looking for. The docs endpoint, /pulp/api/v3/docs/ (view it in a browser), is very helpful for finding what endpoints are available in Pulp.

MichalPysik: Done.

MichalPysik: Fixed.

pulp_file/tests/functional/api/test_crud_content_unit.py
MichalPysik: Fixed.

mdellweg:
Can you add a check to the test using the upload that the upload was deleted after content creation?

MichalPysik:
> Can you add a check to the test using the upload that the upload was deleted after content creation?

I already did that, tt is in the chunked upload test. Or did you mean that I should add it to the first two tests as well?

mdellweg:
> > Can you add a check to the test using the upload that the upload was deleted after content creation?
> 
> I already did that, tt is in the chunked upload test. Or did you mean that I should add it to the first two tests as well?

Right, i'm seeing it now.

pulp_file/tests/functional/api/test_crud_content_unit.py
```diff
@@ -157,3 +209,26 @@ def test_create_file_content_from_chunked_upload(
         task = monitor_task(response.task)
         content = file_content_api_client.read(task.created_resources[0])
         assert content.sha256 == expected_digest
+        # Upload gets deleted if the content gets created
+        with pytest.raises(coreApiException):
+            uploads_api_client.read(upload.pulp_href)
```
mdellweg, line 214: Can you change this test to attempt to create the same content unit twice? I believe in this case our version does not delete the upload object as requested.

pulp_file/app/serializers.py
```diff
@@ -35,21 +35,13 @@ def deferred_validate(self, data):
...
-
-            raise serializers.ValidationError(
-                _(
-                    "There is already a file content with relative path '{path}' and digest "
-                    "'{digest}'."
-                ).format(path=data["relative_path"], digest=data["digest"])
-            )
-
-        return data
+        return content if content.exists() else None
```
mdellweg, line 44:
```suggestion
        return content.get() if content.exists() else None
```
but maybe better:
```suggestion
        return content.first()
```

pulp_file/tests/functional/api/test_crud_content_unit.py
```diff
@@ -151,9 +202,31 @@ def test_create_file_content_from_chunked_upload(
...
+    uploads_api_client.update(
+        upload_href=upload.pulp_href, file=file_2, content_range="bytes 128-255/256"
+    )
+    response = file_content_api_client.create(
+        upload=upload.pulp_href, relative_path=most_recent_path
+    )
+    task = monitor_task(response.task)
+    content = file_content_api_client.read(task.created_resources[0])
+    assert content.sha256 == expected_digest
+    # Upload gets deleted even though no new content got created
```
mdellweg, line 230: :+1:

mdellweg:
Before we can merge this, it needs a bump on the lower bound of the pulpcore dependency. (currently >=3.22.0-dev is expected)

bmbouter:
@MichalPysik I see one comment that needs addressing here (the last one about the dependency bumping). @mdellweg is that right?

MichalPysik:
@mdellweg @bmbouter I changed the required version inside 'requirements.txt', is that it or is there any other file with recommended dependency versions?

requirements.txt
```diff
@@ -1 +1 @@
-pulpcore>=3.21.0.dev,<3.25
+pulpcore>=3.22.0.dev,<3.25
```
bmbouter, line 1: I think it should be `3.22.0` since we won't know to reset this at release time. @mdellweg am I remembering this right?

mdellweg: It would not run in the ci unless we actually _have_ a released 3.22 that's why we need the "dev" here.
We should adjust in the release pipeline.

PR #792, opened 2022-08-08
Moved the range-header tests into download policy tests

dkliban:
Required PR: https://github.com/pulp/pulpcore/pull/3059/

[noissue]

### DIFF
--- a/pulp_file/tests/functional/api/test_download_policies.py
+++ b/pulp_file/tests/functional/api/test_download_policies.py
@@ -1,4 +1,5 @@
 """Tests for Pulp`s download policies."""
+from aiohttp.client_exceptions import ClientResponseError
 import hashlib
 import pytest
 import uuid
@@ -12,7 +13,26 @@ from pulp_smash.pulp3.utils import (
 
 from pulp_file.tests.functional.utils import get_files_in_manifest, download_file
 
+from pulpcore.app import settings
 from pulpcore.client.pulp_file import FileFilePublication, RepositorySyncURL
+
+
+def _do_range_request_download_and_assert(url, range_header, expected_bytes):
+    file1 = download_file(url, headers=range_header)
+    file2 = download_file(url, headers=range_header)
+    assert expected_bytes == len(file1.body)
+    assert expected_bytes == len(file2.body)
+    assert file1.body == file2.body
+
+    assert file1.response_obj.status == 206
+    assert file1.response_obj.status == file2.response_obj.status
+
+    assert str(expected_bytes) == file1.response_obj.headers["Content-Length"]
+    assert str(expected_bytes) == file2.response_obj.headers["Content-Length"]
+
+    assert (
+        file1.response_obj.headers["Content-Range"] == file2.response_obj.headers["Content-Range"]
+    )
 
 
 @pytest.mark.parallel
@@ -25,13 +45,15 @@ def test_download_policy(
     file_repo_api_client,
     file_pub_api_client,
     file_distro_api_client,
-    basic_manifest_path,
+    range_header_manifest_path,
     gen_object_with_cleanup,
     file_content_api_client,
     download_policy,
 ):
     """Test that "on_demand" and "streamed" download policies work as expected."""
-    remote = file_fixture_gen_remote_ssl(manifest_path=basic_manifest_path, policy=download_policy)
+    remote = file_fixture_gen_remote_ssl(
+        manifest_path=range_header_manifest_path, policy=download_policy
+    )
     file_repo = file_repo_api_client.read(file_repo.pulp_href)
     assert file_repo.latest_version_href.endswith("/versions/0/")
 
@@ -59,11 +81,6 @@ def test_download_policy(
     ).to_dict()["results"]
     assert len(content) == len(expected_files)
 
-    # Create a File Publication and assert that the repository_version is set on the Publication.
-    publish_data = FileFilePublication(repository=file_repo.pulp_href)
-    publication = gen_object_with_cleanup(file_pub_api_client, publish_data)
-    assert publication.repository_version is not None
-
     # Create a Distribution
     distribution = gen_object_with_cleanup(
         file_distro_api_client,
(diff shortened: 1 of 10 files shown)


pulp_file/tests/functional/api/test_download_policies.py
```diff
@@ -74,23 +73,105 @@ def test_download_policy(
...
     expected_files_list = list(expected_files)
     content_unit = expected_files_list[0]
     content_unit_url = urljoin(distribution.base_url, content_unit[0])
-    downloaded_file = download_file(content_unit_url)
+    downloaded_file, _ = download_file(content_unit_url)
     actual_checksum = hashlib.sha256(downloaded_file).hexdigest()
     expected_checksum = content_unit[1]
     assert expected_checksum == actual_checksum
 
+    def do_range_request_download(url, range_header, expected_bytes):
```
bmbouter, line 97: I think moving this to the module level would be good. Naming it `_do_range_request_download_and_assert` would make sense to me.

pulp_file/tests/functional/api/test_download_policies.py
```diff
@@ -74,23 +73,105 @@ def test_download_policy(
...
+    range_header = {"Range": "bytes=2042176-3248576"}
+    num_bytes = 1206401
+    content_unit = expected_files_list[3]
+    content_unit_url = urljoin(distribution.base_url, content_unit[0])
+    do_range_request_download(content_unit_url, range_header, num_bytes)
+
+    # Assert that range requests with a negative start value errors as expected
+    content_unit = expected_files_list[4]
+    content_unit_url = urljoin(distribution.base_url, content_unit[0])
+    # The S3 test API project doesn't handle invalid Range values correctly
```
bmbouter, line 143: Is there a link to their tracker about this?

pulp_file/tests/functional/utils.py
```diff
@@ -302,35 +302,35 @@ def get_files_in_manifest(url):
...
     return files
 
 
-def download_file(url, auth=None):
+def download_file(url, auth=None, headers=None):
     """
     Performs a GET request on a URL.
     """
-    return asyncio.run(_download_file(url, auth=auth))
+    return asyncio.run(_download_file(url, auth=auth, headers=headers))
```
bmbouter, line 315: If we're keeping the tuple style response, I think it needs to have a longer+proper docstring.

It would be nice to structure it more, maybe with a DataClass? Just an idea. I'd be satisfied with just a docstring. Is another option just returning the response and would that contain the data (I suspect not).

pulp_file/tests/functional/conftest.py
```diff
@@ -152,6 +152,30 @@ def large_manifest_path(file_fixtures_root):
...
+    file8 = generate_iso(file_fixtures_root.joinpath("range/8.iso"), 4 * one_megabyte)
+
+    generate_manifest(
+        file_fixtures_root.joinpath("range/PULP_MANIFEST"),
+        [file1, file2, file3, file4, file5, file6, file7, file8],
+    )
+    generate_manifest(
+        file_fixtures_root.joinpath("range/PULP_MANIFEST"),
+        [file1, file2, file3, file4, file5, file6, file7, file8],
+    )
```
bmbouter, line 175: Is the intention here to create duplicate entries? If so, a one liner comment would be good.

pulp_file/tests/functional/conftest.py
```diff
@@ -152,6 +152,30 @@ def large_manifest_path(file_fixtures_root):
     return "/large/PULP_MANIFEST"
 
 
+@pytest.fixture
+def range_header_manifest_path(file_fixtures_root):
```
bmbouter, line 156: Adding a docstring here would be nice.

pulp_file/tests/functional/api/test_download_policies.py
dkliban: sounds good to me.

pulp_file/tests/functional/api/test_download_policies.py
dkliban: I haven't looked. I am porting this comment from the existing test that was added here: https://github.com/pulp/pulpcore/commit/df2c8f109ec14e6bdb204058eb575d62fb58ad1a

We should check if this is actually still true. I'll try it out.

pulp_file/tests/functional/utils.py
dkliban: I'll add a docstring. The problem with returning the response object is that the method for reading the content is a coroutine.

pulp_file/tests/functional/conftest.py
dkliban: No, that was not the intent. This code duplication was not intentional.

pulp_file/tests/functional/conftest.py
dkliban: I'll add one here. We should add one to all fixtures actually.

pulp_file/tests/functional/api/test_download_policies.py
dkliban: Both Azure and S3 tests fail for the 'immediate' download policy when the start value of Range header is negative.

pulp_file/tests/functional/utils.py
```diff
@@ -302,35 +302,39 @@ def get_files_in_manifest(url):
...
     return files
 
 
-def download_file(url, auth=None):
-    """
-    Performs a GET request on a URL.
+def download_file(url, auth=None, headers=None):
+    """Download a file.
+
+    :param url: str URL to the file to download
```
bmbouter, line 314: nitpick, but this needs to use the napoleon style docstring per the style guide.

pulp_file/tests/functional/utils.py
bmbouter: What a DataClass object returned? Note that will only work on Python 3.7. I think having it be structured would be good. Not a requirement, but a suggestion.

pulp_file/tests/functional/utils.py
dkliban: This is the style of the rest of the module.

bmbouter (approved):
This looks great. Thanks for making this, and for the various changes during review.

PR #794, opened 2022-08-15
Converts the Distribution CRUD tests to use pytest fixtures

dkliban:
[noissue]

### DIFF
--- a/pulp_file/tests/functional/api/from_pulpcore/test_distributions.py
+++ b/pulp_file/tests/functional/api/from_pulpcore/test_distributions.py
@@ -1,250 +1,164 @@
 """Tests that perform actions over distributions."""
-import unittest
+import pytest
+import json
+from uuid import uuid4
 
-from pulp_smash import api, config, utils
 from pulp_smash.pulp3.bindings import monitor_task
 from pulp_smash.pulp3.utils import (
     gen_distribution,
-    gen_repo,
-    get_content,
-    get_versions,
-    modify_repo,
-    sync,
-)
-from requests.exceptions import HTTPError
-
-from pulp_file.tests.functional.utils import (
-    create_file_publication,
-    gen_file_remote,
-)
-from .constants import (
-    BASE_DISTRIBUTION_PATH,
-    FILE_CONTENT_NAME,
-    FILE_DISTRIBUTION_PATH,
-    FILE_REMOTE_PATH,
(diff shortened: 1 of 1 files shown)


pulp_file/tests/functional/api/from_pulpcore/test_distributions.py
```diff
@@ -11,240 +13,141 @@
...
+):
+    def create_distribution_and_assert(data):
+        with pytest.raises(ApiException) as exc:
+            file_distro_api_client.create(data)
+        assert json.loads(exc.value.body)["base_path"] is not None
+
+    def update_distribution_and_assert(distribution_href, data):
+        with pytest.raises(ApiException) as exc:
+            file_distro_api_client.update(distribution_href, data)
+        assert json.loads(exc.value.body)["base_path"] is not None
```
bmbouter, line 124: I think moving these to the module level and prepending an underscore would be nice.

pulp_file/tests/functional/api/from_pulpcore/test_distributions.py
```diff
@@ -11,240 +13,141 @@
...
+    assert distribution.content_guard is None
+    assert distribution.repository is None
+    assert distribution.publication == publication.pulp_href
+    assert distribution.base_path == distribution_data["base_path"]
+    assert distribution.name == distribution_data["name"]
+
+    # Assert that the publication has a reference to the distribution
+    assert publication.distributions[0] == distribution.pulp_href
+
+    # Test updating name and base_path with 'partial_update'
```
bmbouter, line 72: Breaking this comment into two comments to cover the two blocks would be nice. Also similarly below.

pulp_file/tests/functional/api/from_pulpcore/test_distributions.py
dkliban: I put them in here so I wouldn't have to pass the client object into them. I don't mind moving to module level, but I did want to share the motivation.

dkliban: I updated to using module level methods.

bmbouter: Your explanation gives me closure. Get it!? :)

bmbouter (approved):
This is great. Thank you!

PR #796, opened 2022-08-16
Resolve failures in the nightly performance tests

lubosmj:
[noissue]


lubosmj:
This change effectively resolves the error: `AttributeError: module 'time' has no attribute 'clock'` (https://github.com/pulp/pulp_file/runs/7851596185?check_suite_focus=true#step:12:1686).

mdellweg (approved):

PR #802, opened 2022-08-30
Added test for listing master model remotes

### LINKED ISSUE #800: Test case for listing repositories
Since the new Pulpcore feature https://github.com/pulp/pulpcore/pull/3127 allows users to list all master models of remotes, there should be a test case that covers this functionality.

The scenario should be:
1. Create one or more remotes
2. List them through endpoint /pulp/api/v3/remotes/
3. Make sure the href is the same and maybe other possible checks

MichalPysik:
A new feature has been added to pulpcore, which allows users to list
all master model remotes at /pulp/api/v3/remotes/. A test case
covering this feature has been added.

closes #800

Required-PR: pulp/pulpcore#3150

### DIFF
--- a/pulp_file/tests/functional/api/test_generic_list.py
+++ b/pulp_file/tests/functional/api/test_generic_list.py
@@ -39,3 +39,22 @@ def test_read_all_content_guards_generic(
     assert response.count != 0
     for content_guard in response.results:
         assert content_guard.name is not None
+
+
+@pytest.mark.parallel
+def test_read_all_master_model_remotes_generic(
+    remotes_api_client, gen_object_with_cleanup, file_remote_api_client
+):
+    remote_data = {"name": str(uuid.uuid4()), "url": "http://example.com"}
+    remote1 = gen_object_with_cleanup(file_remote_api_client, remote_data)
+    remote_data = {"name": str(uuid.uuid4()), "url": "http://example.org"}
+    remote2 = gen_object_with_cleanup(file_remote_api_client, remote_data)
+
+    response = remotes_api_client.list()
+    assert response.count != 0
+
+    hrefs = []
+    for remote in response.results:
+        hrefs.append(remote.pulp_href)
+    assert remote1.pulp_href in hrefs
+    assert remote2.pulp_href in hrefs
--- a/pulp_file/tests/functional/conftest.py
+++ b/pulp_file/tests/functional/conftest.py
@@ -32,10 +32,7 @@ _logger = logging.getLogger(__name__)
 def pytest_check_for_leftover_pulp_objects(config):
     file_client = gen_file_client()
 
-    types_to_check = [
-        FileFileAlternateContentSource(file_client),
-        RemotesFileApi(file_client),
-    ]
+    types_to_check = [FileFileAlternateContentSource(file_client)]
     for type_to_check in types_to_check:
         if type_to_check.list().count > 0:
             raise Exception(f"This test left over a {type_to_check}.")


pulp_file/tests/functional/api/test_generic_list.py
```diff
@@ -35,3 +41,24 @@ def test_read_all_content_guards_generic(
...
+@pytest.mark.parallel
+def test_read_all_master_model_remotes_generic(
+    file_remote_api_client, gen_object_with_cleanup, pulpcore_client
+):
+    remote_data = {"name": str(uuid.uuid4()), "url": "http://example.com"}
+    remote1 = gen_object_with_cleanup(file_remote_api_client, remote_data)
+    remote_data = {"name": str(uuid.uuid4()), "url": "http://example.org"}
+    remote2 = gen_object_with_cleanup(file_remote_api_client, remote_data)
+
+    url = "https://" + gethostname() + "/pulp/api/v3/remotes/"
```
mdellweg, line 55: Can we use a fixture providing an instance of `RemotesApi` here?

pulp_file/tests/functional/api/test_generic_list.py
```diff
@@ -35,3 +35,34 @@ def test_read_all_content_guards_generic(
...
+
+    assert r1_found and r2_found
+
+    """url = "https://" + gethostname() + "/pulp/api/v3/remotes/"
+    response = requests.get(url=url)
+    parsed_response = loads(response.text)
+    hrefs = []
+    for result in parsed_response["results"]:
+        hrefs.append(result["pulp_href"])
+    assert remote1.pulp_href in hrefs and remote2.pulp_href in hrefs"""
```
mdellweg, line 68: Maybe you can rephrase the asserts above to this line again. It's way more readable. Actually i would make it two asserts:
```suggestion
    assert remote1.pulp_href in hrefs
    assert remote2.pulp_href in hrefs
```

mdellweg:
Not sure if you should add a misc changelog entry, or just use [noissue] and close the issue by hand.
Preferences, anyone?
@pulp/core

pulp_file/tests/functional/api/test_generic_list.py
```diff
@@ -35,3 +35,34 @@ def test_read_all_content_guards_generic(
...
+@pytest.mark.parallel
+def test_read_all_master_model_remotes_generic(
+    remotes_api_client, gen_object_with_cleanup, file_remote_api_client
+):
+    remote_data = {"name": str(uuid.uuid4()), "url": "http://example.com"}
+    remote1 = gen_object_with_cleanup(file_remote_api_client, remote_data)
+    remote_data = {"name": str(uuid.uuid4()), "url": "http://example.org"}
+    remote2 = gen_object_with_cleanup(file_remote_api_client, remote_data)
+
+    response = remotes_api_client.list()
```
bmbouter, line 49: One issue could be if the system under test has so many remotes that the ones you are looking for are not on the first page. Does this endpoint support `?name=somename`? If it does maybe have your querying use the name to avoid this pagination issue. If it doesn't support pagination, I think we can keep it as-is and we'll address those test failures later.

bmbouter:
> Not sure if you should add a misc changelog entry, or just use [noissue] and close the issue by hand. Preferences, anyone? @pulp/core

I have no preference.

bmbouter:
I'm not sure what else you want to do to this, but it's looking good to me! I'm deferring final LGTM to @mdellweg.

bmbouter:
@mdellweg @MichalPysik FYI I think this line can be removed https://github.com/pulp/pulp_file/blob/main/pulp_file/tests/functional/conftest.py#L37 now that pulpcore is able to assert no-leftovers generally for all Remote types here. If you wanted to do that in this PR I think that would be appropriate.

mdellweg:
> @mdellweg @MichalPysik FYI I think [this line](https://github.com/pulp/pulp_file/blob/main/pulp_file/tests/functional/conftest.py#L37) can be removed now that pulpcore is able to assert no-leftovers generally for all Remote types [here](https://github.com/pulp/pulpcore/blob/main/pulpcore/tests/functional/__init__.py#L65). If you wanted to do that in this PR I think that would be appropriate.

Good catch. Do we check for the leftovers in pulpcore already?

mdellweg (approved):

bmbouter:
> > @mdellweg @MichalPysik FYI I think [this line](https://github.com/pulp/pulp_file/blob/main/pulp_file/tests/functional/conftest.py#L37) can be removed now that pulpcore is able to assert no-leftovers generally for all Remote types [here](https://github.com/pulp/pulpcore/blob/main/pulpcore/tests/functional/__init__.py#L65). If you wanted to do that in this PR I think that would be appropriate.
> 
> Good catch. Do we check for the leftovers in pulpcore already?

We do, it's [here](https://github.com/pulp/pulpcore/blob/main/pulpcore/tests/functional/__init__.py#L65).

PR #803, opened 2022-08-30
Fix generic list tests

mdellweg:
[noissue]

Required-PR: https://github.com/pulp/pulpcore/pull/3150


lubosmj (approved):
Now, we are talking about generic tests...

PR #808, opened 2022-09-12
Add test for hidden distributions in listing

mdellweg:
How do I test an off-by-default-feature?


dralley:
> How do I test an off-by-default-feature?

Good question - I'm not sure.  We could make the test skipped when the feature is disabled, I believe we do that for a few other tests, such as the ones contingent on `ALLOWED_CONTENT_CHECKSUMS`.  At least that would let us run them on a dev system when needed...

dralley:
One pattern I've seen with a few frameworks, and it looks like it may be possible with aiohttp, is that you create an instance of the web application in code (in a *unit test* - not a standalone web server) build a request, ask it to process the request using code API (as opposed to HTTP), and then make assertions on the response.

It looks like that might be possible with aiohttp, but I haven't looked deeply enough at our specific implementation.

.ci/ansible/settings.py.j2
```diff
@@ -43,6 +43,9 @@ AWS_S3_SIGNATURE_VERSION = "s3v4"
 AWS_STORAGE_BUCKET_NAME = "pulp3"
 AWS_S3_ENDPOINT_URL = "http://minio:9000"
 AWS_DEFAULT_ACL = "@none None"
+
+# We need a better place for this...
+HIDE_GUARDED_DISTRIBUTIONS = True
```
mdellweg, line 48: Not sure if we want to add this vi the plugin template.
Should we have a way to specify more options for the different scenarios?

dralley: What do you mean by different scenarios - you mean the more "advanced" option which checks the certificate when serving the index page?

mdellweg: No, i mean we run the test suite in different scenarios (for different storage classes). But we could start changing more settings in those scenarios.

ipanova: I had imagined that this option is off by default and we'd have 2 tests where in one scenario we'd test the results with off by default and in the second tests with the `@override_settings` we'd test that protected repos are hidden.

mdellweg: @override_settings only works for unittests.

mdellweg: It uses this now:
https://github.com/pulp/plugin_template/pull/692

lubosmj:
It might be worth considering https://pytest-django.readthedocs.io/en/latest/configuring_django.html#overriding-individual-settings. Because we can run out of test runners when overriding the settings.

mdellweg:
> It might be worth considering https://pytest-django.readthedocs.io/en/latest/configuring_django.html#overriding-individual-settings. Because we can run out of test runners when overriding the settings.

It's a good tool for unittests. But we don't run the pulp we are testing in the unittest process, so no, it's not going to work.

https://github.com/pulp/pulp_file/pull/808#discussion_r970705269

PR #810, opened 2022-09-15
Add domains compatiability

gerrod3:
Required PR: https://github.com/pulp/pulpcore/pull/3190 
[noissue]


pulp_file/app/models.py
```diff
@@ -37,10 +38,11 @@ class FileContent(Content):
 
     relative_path = models.TextField(null=False)
     digest = models.CharField(max_length=64, null=False)
+    domain = models.ForeignKey("core.Domain", default=get_default_domain, on_delete=models.PROTECT)
```
mdellweg, line 41: Wouldn't we add this on the `Content` model?

[ ] Does not work with uniqueness constraints?

mdellweg: Maybe we can add it as a constraint?

https://docs.djangoproject.com/en/3.2/ref/models/options/#django.db.models.Options.constraints

pulp_file/app/serializers.py
```diff
@@ -39,7 +40,9 @@ def deferred_validate(self, data):
 
     def retrieve(self, validated_data):
         content = FileContent.objects.filter(
-            digest=validated_data["digest"], relative_path=validated_data["relative_path"]
+            digest=validated_data["digest"],
+            relative_path=validated_data["relative_path"],
+            pulp_domain=get_domain(),
```
mdellweg, line 45:
```suggestion
            pulp_domain=get_domain_pk(),
```

pulp_file/tests/functional/api/from_pulpcore/test_content_path.py
```diff
@@ -38,17 +38,20 @@ def test_content_directory_listing(
             gen_distribution(base_path=base_path + path, content_guard=content_guard),
         )
 
-    response = utils.http_get(PULP_CONTENT_BASE_URL).decode("utf-8")
+    base_url = PULP_CONTENT_BASE_URL
+    if settings.DOMAIN_ENABLED:
```
mdellweg, line 42: I think, we have a fixture for retrieving settings.

pulp_file/tests/functional/api/test_domains.py
```diff
@@ -0,0 +1,281 @@
+import pytest
+import uuid
+import json
+
+from pulpcore.app import settings
+from pulpcore.client.pulp_file import ApiException
+from pulpcore.client.pulpcore import ApiException as CoreApiException
+from pulp_smash.pulp3.utils import gen_repo, gen_distribution
```
mdellweg, line 8: Can we try not to need these? What is the gap we have with our current factory fixtures?

PR #812, opened 2022-09-19
Migrate off pulp-smash

### LINKED ISSUE #804: Migrate off of pulp-smash
As determined in [this discourse thread](https://discourse.pulpproject.org/t/retirement-of-pulp-smash/579/2) by the CI/CD mini-team, let's demonstrate one plugin entirely moving off of pulp-smash. pulp_file is a good candidate for that. To do this:

1. Move any necessary pytest fixtures from pulp-smash to pulpcore
2. Remove pulp-smash from the test dependencies
3. Ensure pulp-smash is not being installed at all in the CI either
4. Move any necessary utils functions

It would be good if whatever moves into pulpcore does not have the ability to reach "remote machines under test". 1) We don't want that anymore architecturally and 2) if we do that we have to also move the configuration for such machinery (which we don't want either).

lubosmj:
Required PR: https://github.com/pulp/pulpcore/pull/3223

closes #804


lubosmj:
It has been decided that we first need to rewrite all unittest-based tests to pytest. Then, we can proceed with migrating off pulp-smash.

PR #817, opened 2022-09-21
Update plugin-template

mikedep333:
Includes increasing pulp-workers from 2 to 9

[noissue]


template_config.yml
```diff
@@ -32,6 +32,7 @@ flake8: true
 flake8_ignore: []
 github_org: pulp
 issue_tracker: github
+kanban: true
 lint_requirements: true
 noissue_marker: '[noissue]'
 parallel_test_workers: 8
```
mdellweg, line 38: Is this an old variable?

mikedep333: It got generated when I ran the plugin-template command.

mdellweg: Can we remove `parallel_test_workers` if we are not using it?

dkliban:
Closing because it is out of date.

PR #829, opened 2022-09-23
Add content-disposition header assertions

ipanova:
[noissue]

### DIFF
--- a/CHANGES/828.misc
+++ b/CHANGES/828.misc
@@ -1,0 +1,1 @@
+Added content-disposition header assertions when content is consumed from an object storage.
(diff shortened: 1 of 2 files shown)


bmbouter (approved):

dkliban (approved):
I filed https://github.com/pulp/pulpcore/issues/3255 to capture the lack of `Content-Disposition` header in a few cases.

ipanova:
> I filed [pulp/pulpcore#3255](https://github.com/pulp/pulpcore/issues/3255) to capture the lack of `Content-Disposition` header in a few cases.

Thanks! I was on the fence whether we need an issue in this case, because it takes automatically the relative path from the content_artifact during the download. So unless you want to set content-disposition header filename to something else than relative path than it is redundant.
Plus we need to be careful with this security issue https://www.djangoproject.com/weblog/2022/aug/03/security-releases/

Issue #816, opened 2022-09-20
Remove the test_publications module


dkliban:
The tests in the `test_publications` module[0] need to be removed. Any assertions that are present in `test_publications` but missing in `test_publish` module[1] need to be added to the `test_publish` module. 

[0] https://github.com/pulp/pulp_file/blob/1.11/pulp_file/tests/functional/api/from_pulpcore/test_publications.py
[1] https://github.com/pulp/pulp_file/blob/main/pulp_file/tests/functional/api/test_publish.py

PR #830, opened 2022-09-23
Refactors orphans delete and cleanup tests

dkliban:
These tests now use pytest fixtures.

Required PR: https://github.com/pulp/pulpcore/pull/3250

fixes: #816

### DIFF
--- a/pulp_file/tests/functional/api/from_pulpcore/test_orphans.py
+++ b/pulp_file/tests/functional/api/from_pulpcore/test_orphans.py
@@ -1,300 +1,110 @@
 """Tests that perform actions over orphan files."""
 import os
-import unittest
-from random import choice
+import pytest
 
-from pulp_smash import cli, config, utils
-from pulp_smash.exceptions import CalledProcessError
 from pulp_smash.pulp3.bindings import monitor_task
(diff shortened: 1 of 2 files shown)


pulp_file/tests/functional/api/from_pulpcore/test_orphans.py
```diff
@@ -1,300 +1,103 @@
...
+
+def test_cleanup_specific_orphans(
+    file_content_unit_with_name_factory, file_content_api_client, orphans_cleanup_api_client
+):
+    content_unit_1 = file_content_unit_with_name_factory("1.iso")
+    content_unit_2 = file_content_unit_with_name_factory("2.iso")
+    cleanup_dict = {"content_hrefs": [content_unit_1.pulp_href], "orphan_protection_time": 0}
+    monitor_task(orphans_cleanup_api_client.cleanup(cleanup_dict).task)
+
+    # Assert that content_unit_2 is gone and content_unit_1 is present
```
mdellweg, line 87: Can we also assert that no artifacts will be deleted?

mdellweg (approved):
Nice work. One suggestion, but not mandatory.

PR #838, opened 2022-10-04
Allow publishing without manifest

### LINKED ISSUE #837: Add option to hide PULP_MANIFEST
We have a request to publish some files and we want to use pulp_file. However, these repos won't be ingested by another pulp instance so we'd like to hide the PULP_MANIFEST file. 

I see two ways to do this:

1. Add a publish option to not create the pulp manifest file
2. Add an option (or setting?) that have the content app hide the PULP_MANIFEST file

daviddavis:
fixes #837


pulp_file/app/models.py
```diff
@@ -119,7 +119,7 @@ class FilePublication(Publication, AutoAddObjPermsMixin):
 
     TYPE = "file"
 
-    manifest = models.TextField()
+    manifest = models.TextField(null=True)
```
mdellweg, line 122: I'm wondering if those two appearances of the same field should have the same default?
@dralley, maybe you know.

pulp_file/app/tasks/publishing.py
```diff
@@ -39,11 +39,12 @@ def publish(manifest, repository_version_pk):
     with tempfile.TemporaryDirectory(dir="."):
         with FilePublication.create(repo_version, pass_through=True) as publication:
             publication.manifest = manifest
-            manifest = Manifest(manifest)
-            manifest.write(yield_entries_for_version(repo_version))
-            PublishedMetadata.create_from_file(
-                file=File(open(manifest.relative_path, "rb")), publication=publication
-            )
+            if manifest:
```
mdellweg, line 42: Is auto-publishing taking the same code path?

pulp_file/app/models.py
daviddavis: There’s two ways manifest gets set. The first is from the repo manifest field during auto publish. Then it also can be set during publication creation which has its default here:

https://github.com/pulp/pulp_file/blob/02adbf1efe98a5971abf47a45d0c8ff4025b589f/pulp_file/app/serializers.py#L110

I don’t think it’s strictly necessary to also set a default here but I’m happy to do it.

pulp_file/app/tasks/publishing.py
daviddavis: Yea the auto publish code calls the publish task. Same as the viewset. 

https://github.com/pulp/pulp_file/blob/02adbf1efe98a5971abf47a45d0c8ff4025b589f/pulp_file/app/models.py#L94

mdellweg (approved):

PR #845, opened 2022-10-11
Rewrite test_tasks unittest-based tests to pytest style

### LINKED ISSUE #825: Refactor the test_tasks module to use pytest fixtures
The tests in the `test_tasks` module[0] need to be refactored to use pytest fixtures.

[0] https://github.com/pulp/pulp_file/blob/main/pulp_file/tests/functional/api/from_pulpcore/test_tasks.py

decko:
Rewrites the unittest based tests to pytest style using fixtures.

Closes #825

### DIFF
--- a/pulp_file/tests/functional/api/from_pulpcore/test_tasks.py
+++ b/pulp_file/tests/functional/api/from_pulpcore/test_tasks.py
@@ -1,225 +1,109 @@
 """Test that operations can be performed over tasks."""
-import unittest
+import json
+from urllib.parse import urljoin
+from uuid import uuid4
 
-from pulp_smash import api, config, utils
-from pulp_smash.pulp3.constants import (
-    BASE_DISTRIBUTION_PATH,
-    P3_TASK_END_STATES,
-    TASKS_PATH,
-)
-from pulp_smash.pulp3.utils import gen_repo, gen_distribution, get_content, modify_repo, sync
-from requests import HTTPError
+import pytest
+from aiohttp import BasicAuth
+from pulpcore.tests.functional.utils import monitor_task
+from pulpcore.client.pulp_file import RepositorySyncURL
+from pulpcore.client.pulpcore.exceptions import ApiException
 
-from pulp_file.tests.functional.utils import gen_file_remote, skip_if
-from .constants import (
-    FILE_CONTENT_NAME,
-    FILE_REMOTE_PATH,
-    FILE_REPO_PATH,
-)
+from pulp_file.tests.functional.utils import download_file
 
 
-_DYNAMIC_TASKS_ATTRS = ("finished_at",)
-"""Task attributes that are dynamically set by Pulp, not set by a user."""
+@pytest.fixture
+def distribution(file_repo, file_distro_api_client, gen_object_with_cleanup):
+    distribution = gen_object_with_cleanup(
+        file_distro_api_client,
+        {"name": str(uuid4()), "base_path": str(uuid4()), "repository": file_repo.pulp_href},
+    )
+
+    return distribution
 
 
-class TasksTestCase(unittest.TestCase):
-    """Perform different operation over tasks.
+@pytest.mark.parallel
+def test_retrieve_task_with_fields_created_resources_only(
+    bindings_cfg, tasks_api_client, distribution
+):
+    """Perform filtering over the task's field created_resources."""
 
-    This test targets the following issues:
+    task, *_ = tasks_api_client.list(created_resources=distribution.pulp_href).results
 
-    * `Pulp #3144 <https://pulp.plan.io/issues/3144>`_
-    * `Pulp #3527 <https://pulp.plan.io/issues/3527>`_
-    * `Pulp Smash #754 <https://github.com/pulp/pulp-smash/issues/754>`_
-    """
+    auth = BasicAuth(login=bindings_cfg.username, password=bindings_cfg.password)
+    full_href = urljoin(bindings_cfg.host, task.pulp_href)
 
-    @classmethod
-    def setUpClass(cls):
-        """Create class-wide variables."""
-        cls.client = api.Client(config.get_config(), api.json_handler)
-        cls.task = {}
+    response_body = download_file(f"{full_href}?fields=created_resources", auth=auth).body
 
-    def test_01_create_task(self):
-        """Create a task."""
-        repo = self.client.post(FILE_REPO_PATH, gen_repo())
-        self.addCleanup(self.client.delete, repo["pulp_href"])
-        attrs = {"description": utils.uuid4()}
-        response = self.client.patch(repo["pulp_href"], attrs)
-        self.task.update(self.client.get(response["task"]))
+    filtered_task = json.loads(response_body)
 
-    @skip_if(bool, "task", False)
-    def test_02_read_href(self):
-        """Read a task by its pulp_href."""
-        task = self.client.get(self.task["pulp_href"])
-        for key, val in self.task.items():
-            if key in _DYNAMIC_TASKS_ATTRS:
-                continue
-            with self.subTest(key=key):
-                self.assertEqual(task[key], val, task)
-
-    @skip_if(bool, "task", False)
-    def test_02_read_href_with_specific_fields(self):
-        """Read a task by its pulp_hrefproviding specific fields."""
-        fields = ("pulp_href", "state", "worker")
-        task = self.client.get(self.task["pulp_href"], params={"fields": ",".join(fields)})
-        self.assertEqual(sorted(fields), sorted(task.keys()))
-
-    @skip_if(bool, "task", False)
-    def test_02_read_task_without_specific_fields(self):
-        """Read a task by its href excluding specific fields."""
-        # requests doesn't allow the use of != in parameters.
-        url = "{}?exclude_fields=state".format(self.task["pulp_href"])
-        task = self.client.get(url)
-        self.assertNotIn("state", task.keys())
-
-    @skip_if(bool, "task", False)
-    def test_02_read_task_with_minimal_fields(self):
-        """Read a task by its href filtering minimal fields."""
-        task = self.client.get(self.task["pulp_href"], params={"minimal": True})
-        response_fields = task.keys()
-        self.assertNotIn("progress_reports", response_fields)
-        self.assertNotIn("parent_task", response_fields)
-        self.assertNotIn("error", response_fields)
-
-    @skip_if(bool, "task", False)
-    def test_02_read_invalid_worker(self):
-        """Read a task using an invalid worker name."""
-        with self.assertRaises(HTTPError):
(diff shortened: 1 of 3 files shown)


pulp_file/tests/functional/api/from_pulpcore/test_tasks.py
```diff
@@ -22,6 +28,75 @@
 """Task attributes that are dynamically set by Pulp, not set by a user."""
 
 
+@pytest.fixture
```
dkliban, line 31: We only need one task for all the tests. Let's set the scope to session. `@pytest.fixture(scope="session")`

pulp_file/tests/functional/api/from_pulpcore/test_tasks.py
```diff
@@ -22,6 +28,75 @@
...
+
+
+def test_retriving_limited_task_fields(task, config):
+    """Verify for specific fields retrieve in the payload."""
+    expected_fields = set(("pulp_href", "state", "worker"))
+
+    auth = BasicAuth(login=config.username, password=config.password)
+    full_href = urljoin(config.host, task.pulp_href)
+
+    response = download_file(f"{full_href}?fields={','.join(expected_fields)}", auth=auth)
```
dkliban, line 52: Earlier today I saw a different approach for doing a GET request using the API Client instead of the 'download_file' method. It seems a bit more elegant than what I share with you. It's using urllib3 to do the request, so the request() method accepts a dictionary of GET parameters. 

https://github.com/pulp/pulpcore/blob/6857606f5d6136281747317a5699ebaa08cb1699/pulpcore/tests/functional/api/test_api_docs.py#L17

dkliban:
@decko This looks really good. I left a couple of comments inline.

pulp_file/tests/functional/api/from_pulpcore/test_tasks.py
decko: I tried to close the scope not only of this fixture, but also an expensive one we need by the end of the testing module, but always get this message:
```
ScopeMismatch: You tried to access the function scoped fixture file_repo with a module scoped request object, involved factories:
pulp_file/tests/functional/api/from_pulpcore/test_tasks.py:22:  def task(file_repo, file_repo_api_client, tasks_api_client)
pulp_file/tests/functional/conftest.py:97:  def file_repo(file_repo_api_client, gen_object_with_cleanup)
```
Don't know how, 'cause the `file_repo` fixture is not explicitly scoped. https://github.com/pulp/pulp_file/blob/main/pulp_file/tests/functional/conftest.py#L97-L99

mdellweg: Right. I understand. First "not explicitely scoped" is the same as "scope=test". Once you use a fixture in another fixture the former needs to have at least as tight a scope as the latter one.
All our "api_client" fixtures depend on "cid" (to make the calls in the logs relatable to a failed test) and are therefore test scoped. Maybe not the best design after all.

dkliban: We could do without a fixture for a `task` if we just combined all the tests that deal with the task into one test.

pulp_file/tests/functional/api/from_pulpcore/test_tasks.py
```diff
@@ -1,225 +1,276 @@
...
+@pytest.fixture
+def config(file_repo_api_client):
+    return file_repo_api_client.api_client.configuration
+
+
+"""Perform different operation over tasks.
+
+This test targets the following issues:
+
+* `Pulp #3144 <https://pulp.plan.io/issues/3144>`_
```
dkliban, line 52: We don't need to leave these old references to issues. Let's just focus the comments on what the test is doing.

pulp_file/tests/functional/api/from_pulpcore/test_tasks.py
decko: :thinking:

pulp_file/tests/functional/api/from_pulpcore/test_tasks.py
decko: Remove those docstrings.

pulp_file/tests/functional/api/from_pulpcore/test_tasks.py
decko: I used a [PR](https://github.com/pulp/pulpcore/pull/3314/files#) from @mdellweg as example and created a `cheap task` object to be used. Also, I've set all the tests that used that fixture to be run in parallel.
Yet, I merged the tests that needed the expensive `setup_filter_fixture` making it run just one time.

pulp_file/tests/functional/api/from_pulpcore/test_tasks.py
```diff
@@ -1,225 +1,244 @@
...
+):
+
+    remote = file_fixture_gen_remote_ssl(manifest_path=basic_manifest_path, policy="immediate")
+
+    body = RepositorySyncURL(remote=remote.pulp_href)
+    synced_repo = monitor_task(file_repo_api_client.sync(file_repo.pulp_href, body).task)
+
+    file_repo = file_repo_api_client.read(file_repo.pulp_href)
+
+    for file_content in get_content(file_repo.to_dict())[FILE_CONTENT_NAME]:
```
dkliban, line 195: We are trying to move away from using pulp-smash. Here is how you can create a repository version by adding a random content unit[0].

[0] https://github.com/pulp/pulp_file/blob/main/pulp_file/tests/functional/api/test_auto_publish.py#L70-L76

pulp_file/tests/functional/api/from_pulpcore/test_tasks.py
```diff
@@ -1,225 +1,244 @@
...
         )
 
+    repo_update_action = file_repo_api_client.partial_update(
+        file_repo.pulp_href, {"description": str(uuid4())}
+    )
+    repo_update_task = tasks_api_client.read(repo_update_action.task)
+
+    yield (file_repo, remote, synced_repo, repo_update_task)
+
+    file_repo_api_client.delete(file_repo.pulp_href)
```
dkliban, line 207: You don't need to schedule the deletion of file_repo because the fixture already does it after the test runs. The same for the remote.

pulp_file/tests/functional/conftest.py
```diff
@@ -350,3 +347,26 @@ async def handler(request):
 @pytest.fixture
 def bad_response_fixture_server(file_fixtures_root, gen_bad_response_fixture_server):
     yield gen_bad_response_fixture_server(file_fixtures_root, None)
+
+
+@pytest.fixture
+def dispatch_task(cid):
```
dkliban, line 353: Since this fixture is already available in `pulpcore` we should just use it from there[0]. It might not be available right now because it's declared in a test module instead of `conftest` module[1]. @mdellweg am I understanding correctly?

[0] https://github.com/pulp/pulpcore/blob/main/pulpcore/tests/functional/api/test_tasking.py#L12
[1] https://github.com/pulp/pulpcore/blob/main/pulpcore/tests/functional/__init__.py

decko: Can we leave a copy in `pulp_file` and create an issue to come back here later when we add this to pulpcore's `conftest`?

pulp_file/tests/functional/api/from_pulpcore/test_tasks.py
```diff
@@ -1,225 +1,244 @@
...
+    response = tasks_api_client.list(started_at=task.started_at)
+
+    assert response.results and response.count
+
+
+@pytest.mark.parallel
+def test_search_task_by_name(task, tasks_api_client):
+    assert task.name
+
+    search_results = tasks_api_client.list(name=task.name).results
```
mdellweg, line 133: This is a bit dangerous if we start the same task more than pagesize times.
Maybe check that the list is not empty, and all the entries have that name.

pulp_file/tests/functional/api/from_pulpcore/test_tasks.py
mdellweg: We should be able to use the bindings with the "fields" parameter too.

pulp_file/tests/functional/api/from_pulpcore/test_tasks.py
```diff
@@ -1,225 +1,244 @@
...
+        file_distro_api_client,
+        {"name": str(uuid4()), "base_path": str(uuid4()), "repository": file_repo.pulp_href},
+    )
+
+    return distribution
+
+
+@pytest.fixture
+def config(file_repo_api_client):
+    return file_repo_api_client.api_client.configuration
```
mdellweg, line 41: Is this different from:

https://github.com/pulp/pulp-smash/blob/main/pulp_smash/pulp3/pytest_plugin/__init__.py#L310 ?

decko: > We should be able to use the bindings with the "fields" parameter too.

I believe we can't @mdellweg. When pairing with @dkliban we found an unexpected behavior when using the `fields` parameter, where some fields are expected to be in the payload, like `name` and `logging_cid`.
I'm about to test to regenerate the bindings using a new version of the `openapi-generator` to see if this behavior continues or how we're gonna deal with it.

pulp_file/tests/functional/api/from_pulpcore/test_tasks.py
dkliban: The bindings with the fields parameters doesn't work because the model that is instantiated with the response complains that some fields are not set.

pulp_file/tests/functional/api/from_pulpcore/test_tasks.py
decko: https://github.com/pulp/pulp_file/pull/845#discussion_r998332644

pulp_file/tests/functional/api/from_pulpcore/test_tasks.py
```diff
@@ -1,225 +1,109 @@
...
+    repo_update_task = tasks_api_client.read(repo_update_action.task)
+
+    yield (file_repo, remote, synced_repo, repo_update_task)
+
+    tasks_api_client.delete(repo_update_task.pulp_href)
+
+
+def test_filter_tasks_by_reserved_resources(setup_filter_fixture, tasks_api_client):
+    """Filter all tasks by a particular reserved resource."""
+    *_, synced_repo, repo_update_task = setup_filter_fixture
```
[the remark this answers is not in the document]
decko: I believe we could do that, but then we "hit" the wall of ` dispatch_task` fixture not being available outside of pulpcore's `test_tasking` yet.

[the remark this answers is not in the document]
mdellweg: Well either we make the fixture more broadly available or we move this test there too.
I'm fine with both approaches.

dkliban (approved):
Thank you!

PR #849, opened 2022-10-18
Update GHA configs with plugin template

dkliban:
[noissue]


bmbouter (approved):
Looks right to me. Thanks!

PR #855, opened 2022-10-27
Rewrite export tests to pytest

### LINKED ISSUE #818: Refactor the test_pulpexport module to use pytest fixtures
The `test_pulpexport` module[0] needs to be refactored to use pytest fixtures.

[0] https://github.com/pulp/pulp_file/blob/main/pulp_file/tests/functional/api/from_pulpcore/test_pulpexport.py

mdellweg:
fixes #818


pulp_file/tests/functional/api/test_pulp_export.py
```diff
@@ -0,0 +1,282 @@
...
+
+
+@pytest.fixture
+def pulp_exporter_factory(
+    exporters_pulp_api_client,
+    gen_object_with_cleanup,
+    add_to_filesystem_cleanup,
+):
+    def _pulp_exporter_factory(repositories=None):
+        if repositories is None:
```
dkliban, line 28: Maybe just change the default to `[]`?

mdellweg: Rule of python: Do not put mutable objects in function signatures!

dralley: Indeed, it's a major footgun.

pulp_file/tests/functional/conftest.py
```diff
@@ -8,6 +8,7 @@
 import pytest
 from aiohttp import web
 from pulp_smash.pulp3.utils import gen_repo
+from pulp_smash.pulp3.bindings import monitor_task
```
dkliban, line 11: I believe `pulpcore.tests` has `monitor_task` also.

pulp_file/tests/functional/conftest.py
```diff
@@ -232,6 +230,10 @@ def _file_fixture_gen_remote(*, manifest_path, policy, **kwargs):
     yield _file_fixture_gen_remote
 
 
+# TODO rename the original fixture
```
dkliban, line 233: Why do you want to change the name to `file_remote_factory` without actually renaming it?

pulp_file/tests/functional/api/test_pulp_export.py
```diff
@@ -0,0 +1,282 @@
...
+    # UPDATE
+    body = {"path": "/tmp/{}".format(str(uuid.uuid4()))}
+    result = exporters_pulp_api_client.partial_update(exporter.pulp_href, body)
+    monitor_task(result.task)
+    exporter_read = exporters_pulp_api_client.read(exporter.pulp_href)
+    assert exporter_read.path != exporter.path
+    assert exporter_read.path == body["path"]
+
+    # LIST
+    exporters = exporters_pulp_api_client.list().results
```
dkliban, line 144: I tend to filter the results in list assertions. The test would fail if there were a lot of exporters and this one was not on the first page of the results. Seems improbably, but the test does run in parallel. Could possibly fail if run against a real system.

dkliban (changes requested):
Some questions and suggested changes inline. Overall this looks amazing!

pulp_file/tests/functional/api/test_pulp_export.py
dkliban: thanks for the explanation. i was not aware of this.

pulp_file/tests/functional/conftest.py
mdellweg: Too many `monitor_task` implementations around.
I intend to provide a fixture for it here: https://github.com/pulp/pulpcore/pull/3378
Can we just consolidate that later?

pulp_file/tests/functional/conftest.py
mdellweg: I'm going to pull back on the renaming for this pr.

mdellweg:
Incorporated: https://github.com/pulp/pulp_file/pull/856

pulp_file/tests/functional/api/test_pulp_export.py
```diff
@@ -0,0 +1,285 @@
...
+
+
+@pytest.fixture(autouse=True)
+def check_allowed_paths(pulp_cfg):
+    allowed_export_paths = settings.ALLOWED_EXPORT_PATHS
+    if "/tmp" not in allowed_export_paths:
+        pytest.skip(
+            "Cannot run export-tests unless /tmp is in ALLOWED_EXPORT_PATHS "
+            f"({allowed_export_paths})."
+        )
```
lubosmj, line 18: I am not sure if this will work for tests that do not contain the `pulp_cfg` fixture. A safer way for me is doing this: https://github.com/pulp/pulp_file/blob/6651e7988f37072688bb6445ceda230755d7770e/pulp_file/tests/functional/api/from_pulpcore/test_repair.py#L21-L27.

What do you think?

pulp_file/tests/functional/utils.py
```diff
@@ -186,67 +137,11 @@ def delete_exporter(exporter):
     exporter_api = ExportersPulpApi(core_client)
     cmd = ("rm", "-rf", exporter.path)
 
-    cli_client.run(cmd, sudo=True)
+    cli_client.run(cmd)
```
dkliban, line 140: you don't need to use cli_client anymore. Just use `os.rmdir` to remove the directory.

pulp_file/tests/functional/api/test_pulp_export.py
```diff
@@ -0,0 +1,285 @@
...
+@pytest.fixture
+def repository_with_four_versions(
+    file_repo_api_client,
+    file_fixture_gen_file_repo,
+    file_content_api_client,
+    random_artifact,
+):
+    repository = file_fixture_gen_file_repo()
+    for i in range(3):
+        file_content_api_client.create(
```
lubosmj, line 107: The response for such a request should be a task (https://docs.pulpproject.org/pulp_file/restapi.html#tag/Content:-Files/operation/content_file_files_create). Did you forget to call `monitor_task`?

pulp_file/tests/functional/utils.py
mdellweg: Yes!

pulp_file/tests/functional/api/test_pulp_export.py
mdellweg: Good catch!

pulp_file/tests/functional/api/test_pulp_export.py
lubosmj: Is it possible to add the `allow_module_level=True` parameter to the `skip` method?

mdellweg: Changed to module level skip.

dkliban (approved):

PR #858, opened 2022-11-18
Rework api_client fixture usage

mdellweg:
[noissue]

### DIFF
--- a/pulp_file/tests/functional/conftest.py
+++ b/pulp_file/tests/functional/conftest.py
@@ -8,7 +8,6 @@ import aiofiles
 import pytest
 from aiohttp import web
 from pulp_smash.pulp3.utils import gen_repo
-from pulp_smash.pulp3.bindings import monitor_task
 from pulpcore.client.pulp_file import (
     AcsFileApi,
     ApiClient,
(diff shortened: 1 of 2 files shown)


pulp_file/tests/functional/conftest.py
```diff
@@ -35,59 +34,71 @@ def pytest_check_for_leftover_pulp_objects(config):
...
+@pytest.fixture(scope="session")
+def file_remote_api_client(file_client):
+    return RemotesFileApi(file_client)
+
+
+# Factory fixtures
+
+
+@pytest.fixture
+def file_random_content_unit(file_content_unit_with_name_factory):
```
dkliban, line 87: I thought we already had a `file_random_content_unit` fixture but it seems like this is a new fixture that is not used anywhere yet. Is that correct?

mdellweg: This one only moved around, because i wanted all the api_clients to cuddle in one place. I did not check whether it is in use.

dkliban: I now see that you moved it around. I wasn't seeing earlier.

dkliban (approved):

PR #860, opened 2022-11-30
Removed a test for labels in urlencoded post data

mdellweg:
It turns out there is no proper documentation about how it is done.

[noissue]


mdellweg:
This is needed for:
https://github.com/pulp/pulpcore/pull/3427

dkliban (approved):

PR #863, opened 2022-12-02
Adds test assertions for dates on HTML pages of pulpcore-content

dkliban:
closes: #862

Required PR: https://github.com/pulp/pulpcore/pull/3431

### DIFF
--- a/pulp_file/tests/functional/api/test_download_policies.py
+++ b/pulp_file/tests/functional/api/test_download_policies.py
@@ -1,8 +1,10 @@
 """Tests for Pulp`s download policies."""
 from aiohttp.client_exceptions import ClientResponseError
+from bs4 import BeautifulSoup
 import hashlib
 import os
 import pytest
+import subprocess
 import uuid
 from urllib.parse import urljoin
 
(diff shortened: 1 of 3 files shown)


pulp_file/tests/functional/api/test_download_policies.py
```diff
@@ -110,6 +111,17 @@ def test_download_policy(
...
+    # Assert that the dates on the distribution listing page represent the date that the content
+    # was created in Pulp
+    zero_dot_iso = file_content_api_client.list(
+        repository_version=file_repo.latest_version_href, relative_path="foo/0.iso"
+    ).results[0]
+    distribution_html_page = download_file(f"{distribution.base_url}foo/")
+    soup = BeautifulSoup(distribution_html_page.body, "html.parser")
+    all_strings = [s for s in soup.strings if s != "\n"]
+    assert all_strings[3] == "0.iso"
+    assert all_strings[4].strip() == zero_dot_iso.pulp_created.strftime("%d-%b-%Y %H:%M")
```
daviddavis, line 123: There is a slight chance since you are comparing the time the content was created to the time the CA was created that the time won't match. Imagine that one was created at 10:00:01 and the other was created at 9:59:58 for example. I don't know if you want to worry about it but you could use `datetime.timedelta`:

```python
from datetime import timedelta

ca_created = datetime.strptime(all_strings[4].strip(), "%d-%b-%Y %H:%M")
content_created = datetime.strptime(zero_dot_iso.pulp_created, "%d-%b-%Y %H:%M")
assert ca_created - content_created < timedelta(seconds=30)
```

dkliban: We now assume the tests are running on the same machine as Pulp. So I could just look up the content artifact and compare it's date. I'll do that.

daviddavis (approved):

bmbouter (approved):
Very nice!

PR #872, opened 2022-12-14
Removed a test for labels in urlencoded post data

mdellweg:
It turns out there is no proper documentation about how it is done.

[noissue]

(cherry picked from commit 167497d289d87d4bdfc4ab2008666bd0505e454f)


mdellweg:
This is needed to unblock pulp_file nightlies.

ggainey (approved):

PR #873, opened 2022-12-14
Revert "Adds tests for 301 redirect responses from pulpcore-content"

lubosmj:
Reverts pulp/pulp_file#865

This is required to make the CI green again after addressing https://github.com/pulp/pulpcore/pull/3465.


mdellweg:
Should we rather skip the tests?

ggainey (approved):

PR #892, opened 2023-02-03
Skipping test_auto_publish_and_distribute, relies not-yet-released core.

ggainey:
[noissue]


mdellweg (approved):
Maybe specify the version of pulpcore? But then again, it's only on a release branch...

ggainey:
> Maybe specify the version of pulpcore? But then again, it's only on a release branch...

Excellent point!

PR #894, opened 2023-02-06
Test validation exposed by the repository modify endpoint

lubosmj:
Required PR: https://github.com/pulp/pulpcore/pull/3552

[noissue]

### DIFF
--- a/pulp_file/tests/functional/api/test_crud_content_unit.py
+++ b/pulp_file/tests/functional/api/test_crud_content_unit.py
@@ -8,6 +8,7 @@ from pulpcore.tests.functional.utils import PulpTaskError
 from pulp_smash.pulp3.utils import get_content_summary, get_added_content_summary
 
 from pulpcore.client.pulpcore.exceptions import ApiException as coreApiException
+from pulpcore.client.pulp_file.exceptions import ApiException
 
 
 @pytest.mark.parallel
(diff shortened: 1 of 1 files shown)


pulp_file/tests/functional/api/test_crud_content_unit.py
```diff
@@ -180,6 +181,14 @@ def test_cannot_create_repo_version_with_two_relative_paths_the_same(
         monitor_task(response.task)
 
 
+@pytest.mark.parallel
+def test_bad_inputs_to_modify_endpoint(file_repo, file_repo_api_client, file_random_content_unit):
+    body = {"add_content_units": file_random_content_unit}
```
mdellweg, line 186: This is testing with a dict instead of a list of strings, right?
How about 
```suggestion
    body = {"add_content_units": {}}
```
Maybe also test:
- `[{}]`
- `["/content/"]`
- `[f"{api_base_path}/contents/file/file/abcd"]`

lubosmj: Great suggestion, thanks!

mdellweg (commented):
About to approve, but we probably cannot merge until pulpcore is released, right?
Also does this need a pulpcore req version bump, or should we make the test conditionally skipped?

mdellweg:
> This test verifies the fix that is going to be released. I can do this to conditionally skip the test:
> 
> ```python
> import pkg_resources
> 
> PULPCORE_VERSION  = pkg_resources.get_distribution("pulpcore").version
> 
> ...
> 
> def test_XXX():
>    if PULPCORE_VERSION < X.Y.Z:
>       pytest.skip("Blabla")
> ```

Want to have more bye in from @pulp/core here.

pulp_file/tests/functional/api/test_crud_content_unit.py
```diff
@@ -180,6 +181,20 @@ def test_cannot_create_repo_version_with_two_relative_paths_the_same(
         monitor_task(response.task)
 
 
+@pytest.mark.parallel
+def test_bad_inputs_to_modify_endpoint(file_repo, file_repo_api_client, has_pulp_plugin):
+    has_pulp_plugin("core", min="3.23.0.dev")
```
mdellweg, line 186:
```suggestion
    needs_pulp_plugin("core", min="3.23.0.dev")
```
If you want the conditional skip.

lubosmj: Oh, jeez, sorry.

mdellweg (approved):

PR #895, opened 2023-02-10
Fixed test_download_policy behavior when REDIRECT_TO_OBJECT_STORAGE=f…

MichalPysik:
…alse

The download policy test had been checking for the 'Content-Disposition' header when REDIRECT_TO_OBJECT_STORAGE was set to false, in which case it would fail. This commit adds additional conditional check.

[noissue]

### DIFF
--- a/pulp_file/tests/functional/api/test_download_policies.py
+++ b/pulp_file/tests/functional/api/test_download_policies.py
@@ -139,7 +139,11 @@ def test_download_policy(
     actual_checksum = hashlib.sha256(downloaded_file.body).hexdigest()
     expected_checksum = content_unit[1]
     assert expected_checksum == actual_checksum
-    if download_policy == "immediate" and settings.DEFAULT_FILE_STORAGE in OBJECT_STORAGES:
+    if (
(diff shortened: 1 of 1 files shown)


pulp_file/tests/functional/api/test_download_policies.py
```diff
@@ -139,7 +139,11 @@ def test_download_policy(
     actual_checksum = hashlib.sha256(downloaded_file.body).hexdigest()
     expected_checksum = content_unit[1]
     assert expected_checksum == actual_checksum
-    if download_policy == "immediate" and settings.DEFAULT_FILE_STORAGE in OBJECT_STORAGES:
+    if (
+        download_policy == "immediate"
+        and settings.DEFAULT_FILE_STORAGE in OBJECT_STORAGES
```
mdellweg, line 144: From the logic of the settings, I think this should be
```suggestion
        and settings.DEFAULT_FILE_STORAGE != "pulpcore.app.models.storage.FileSystem"
```

mdellweg (approved):

Issue #909, opened 2023-03-06
Add ability to repicate file repositories from an upstream Pulp


dkliban:
pulpcore 3.23 provides a Replicator class in the plugin API. The file plugin should provide its implementation of the Replicator so the replicate API can discover File Distributions to sync from an upstream Pulp.

PR #910, opened 2023-03-06
Adds PulpFileReplicator

dkliban:
closes: #909

### DIFF
--- a/CHANGES/909.feature
+++ b/CHANGES/909.feature
@@ -1,0 +1,1 @@
+Added the ability to replicate File distributions/repositories from an upstream Pulp instance.
--- a/pulp_file/app/replica.py
+++ b/pulp_file/app/replica.py
@@ -1,0 +1,53 @@
+from pulpcore.plugin.replica import Replicator
+
+from pulp_glue.file.context import (
+    PulpFileDistributionContext,
+    PulpFilePublicationContext,
+    PulpFileRepositoryContext,
+)
+
+from pulp_file.app.models import FileDistribution, FileRemote, FileRepository
+from pulp_file.app.tasks import synchronize as file_synchronize
+
+
+class FileReplicator(Replicator):
+    repository_ctx_cls = PulpFileRepositoryContext
+    distribution_ctx_cls = PulpFileDistributionContext
+    publication_ctx_cls = PulpFilePublicationContext
+    app_label = "file"
+    remote_model_cls = FileRemote
+    repository_model_cls = FileRepository
+    distribution_model_cls = FileDistribution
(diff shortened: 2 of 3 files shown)


CHANGES/909.feature
```diff
@@ -0,0 +1 @@
+Adds ability to replicate File distributions/repositories from an upstream Pulp.
```
lubosmj, line 1:
```suggestion
Added the ability to replicate File distributions/repositories from an upstream Pulp instance.
```

dkliban: Thanks for the suggestion! I updated the changelog entry.

pulp_file/app/replica.py
```diff
@@ -0,0 +1,53 @@
...
+    remote_model_cls = FileRemote
+    repository_model_cls = FileRepository
+    distribution_model_cls = FileDistribution
+    distribution_serializer_name = "FileDistributionSerializer"
+    repository_serializer_name = "FileRepositorySerializer"
+    remote_serializer_name = "FileRemoteSerializer"
+    sync_task = file_synchronize
+
+    def url(self, upstream_distribution):
+        # Check if a distribution is repository or publication based
```
ipanova, line 27: what about distribution which is repository_version based?

pulp_file/app/replica.py
```diff
@@ -0,0 +1,53 @@
...
+                self.pulp_ctx, upstream_distribution["publication"]
+            ).entity["manifest"]
+        else:
+            # This distribution doesn't serve any content
+            return None
+
+        return f"{upstream_distribution['base_url']}{manifest}"
+
+    def repository_extra_fields(self, remote):
+        return dict(manifest=remote.url.split("/")[-1], autopublish=True)
```
mdellweg, line 43: Just a matter of style...
```suggestion
        return {"manifest": remote.url.split("/")[-1], "autopublish": True}
```

pulp_file/app/replica.py
```diff
@@ -0,0 +1,53 @@
...
+
+    def sync_params(self, repository, remote):
+        return dict(
+            remote_pk=str(remote.pk),
+            repository_pk=str(repository.pk),
+            mirror=True,
+        )
```
mdellweg, line 50: Same:
```suggestion
    def sync_params(self, repository, remote):
        return {
            "remote_pk": str(remote.pk),
            "repository_pk": str(repository.pk),
            "mirror": True,
        }
```

pulp_file/app/replica.py
ipanova: Thanks for confirmation!

ipanova (approved):

PR #919, opened 2023-03-23
Remove client bindings from functest_requirements

mdellweg:
They should be there for consistency, but this breaks with older branches.

[noissue]

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -18,6 +18,4 @@ values =
 
 [bumpversion:file:./setup.py]
 
-[bumpversion:file:./functest_requirements.txt]
-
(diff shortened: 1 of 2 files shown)


mdellweg:
But we need to do something. What do folks prefer?

@lubosmj @ggainey

lubosmj (approved):
I do not have any strong opinion on this. We should at least make sure there are no discrepancies between Pulp repositories in terms of having different requirements when it comes to bindings (in pulp_file, we require bindings, but not in pulp_container). I find it a bit useful that we state here that in order to run the functional tests locally, we need to install the bindings first.

mdellweg:
The matter of fact is we have it absolutely not consistent across plugins.

PR #924, opened 2023-03-26
Update CI files from 1.10 branch

pulpbot:
[noissue]


requirements.txt
```diff
@@ -1 +1 @@
-pulpcore>=3.16.0,<3.25
+pulpcore>=3.17.0,<3.25
```
ggainey, line 1: This is the problem, right here - file/1.10 fails the acs-refresh-call with core/3.17 (and not with core/3.16, which was its original lb)

ggainey (dismissed):

requirements.txt
mdellweg: I tried bumping that because it failed with 3.16.0 in very much the same way.

mdellweg (approved):

PR #943, opened 2023-04-05
Fix test for file size in content app directory

gerrod3:
Related PR: https://github.com/pulp/pulpcore/pull/3696

### DIFF
--- a/pulp_file/tests/functional/api/test_download_policies.py
+++ b/pulp_file/tests/functional/api/test_download_policies.py
@@ -58,6 +58,7 @@ def test_download_policy(
     file_content_api_client,
     monitor_task,
     pulp_settings,
+    has_pulp_plugin,
     download_policy,
 ):
     """Test that "on_demand" and "streamed" download policies work as expected."""
@@ -107,7 +108,7 @@ def test_download_policy(
     # Assert that un-published content is not available
     for expected_file in expected_files:
         with pytest.raises(ClientResponseError) as exc:
-            content_unit_url = urljoin(distribution.base_url, expected_file[1])
+            content_unit_url = urljoin(distribution.base_url, expected_file[0])
             download_file(content_unit_url)
         assert exc.value.code == 404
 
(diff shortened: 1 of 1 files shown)


pulp_file/tests/functional/api/test_download_policies.py
```diff
@@ -106,8 +108,10 @@ def test_download_policy(
 
     # Assert that un-published content is not available
     for expected_file in expected_files:
+        if "0.iso" in expected_file[0]:
+            artifact_size = humanize.naturalsize(expected_file[2])
```
decko, line 112: Since we're using just this simple function, what if we just copy(and add a reference to the original) the `naturalsize` function into our code base instead of adding the complete library?

gerrod3: Here is the source code, do you think it's small enough to copy to our utils? https://github.com/python-humanize/humanize/blob/main/src/humanize/filesize.py

lubosmj: We were always cautious about adding new dependencies. Let's invite more @pulp/core developers into the discussion.

ipanova: This is a new dep to the functional tests not to the pulp itself. I am ok with that.

decko: > Here is the source code, do you think it's small enough to copy to our utils? https://github.com/python-humanize/humanize/blob/main/src/humanize/filesize.

If you don't count the `gnu` and `binary` related code and remove the comments, it's gonna use less than 30 lines.

gerrod3: My new change in pulpcore adds file size to the content app listing, but I format the size through jinja2's filesizeformat filter in the template: https://jinja.palletsprojects.com/en/3.1.x/templates/#jinja-filters.filesizeformat. This filter isn't exposed for import so I had to use another library to get a similar format style.

ggainey: > This filter isn't exposed for import so I had to use another library to get a similar format style

Hm, that makes sense. It does mean that we're not testing "the same thing" - if humanize or jinja.filesizeformat ever, for example, change the default to binary=True, then suddenly this testcase will fail. Not a huge deal perhaps, but def the problem with relying on different code to do "the same thing".

Can we do a looser test using "contains", both here and where we're using artifact_size, below?

mdellweg: We do control the artifacts created here. So we should know their size exactly. Can we test against a literal string please?

ggainey (approved):

mdellweg (approved):

PR #945, opened 2023-04-16
Update CI files from main branch

pulpbot:
[noissue]


lubosmj (approved):
Seems to resolve all the failures.

mdellweg:
Also i'm unsure if it is going in the right direction. I kind of feel, we should stop installing pulpcore's doc_requirements at all.

PR #955, opened 2023-05-02
Push max-compat to core/3.40.

### LINKED ISSUE #954: Update upper-compat to core/3.40
With the release of core/3.24, there needs to be a version of pulp_file that is compatible with core/3.25 in order for core/3.25.dev to have test-availability. We need a pulp_file release that bumps the upper-bound.

ggainey:
fixes #954.


lubosmj (dismissed):

ggainey:
"It seems like more changes are required because we are failing on domains."

Yeah, clearly. These tests were passing against core/3.23, something about 3.24 (I believe) is breaking them. Will pick up investigating tomorrow.

ggainey:
Now that we've fixed task-resources and released core/3.24.1, this runs clean

lubosmj (approved):

PR #959, opened 2023-05-11
Added migration necessitated by Django4.2/core--3.25 update.

ggainey:
[noissue]


pulp_file/app/migrations/0017_alter_filealternatecontentsource_alternatecontentsource_ptr_and_more.py
```diff
@@ -0,0 +1,51 @@
...
+
+from django.db import migrations, models
+import django.db.models.deletion
+import pulpcore.app.util
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('core', '0106_alter_artifactdistribution_distribution_ptr_and_more'),
```
mdellweg, line 11: So this line is the reason, why we depend hard on 3.25 here, right? I guess, I'm not sufficiently confident to say, scrape it off.

ggainey: Hm, that's a good point. If we can/could do so safely, it would make our release-dance today WAY simpler(ish)

ggainey: Once github has calmed down, let me repush w/ the old lb and that experiment, and we can see how the tests go

ggainey: OK, poulling the core_ requirement "seems to work" - eyes please?

mdellweg (approved):

PR #969, opened 2023-06-13
Adds test for re-import after broken import-attempt.

ggainey:
Tests the support added to fix https://github.com/pulp/pulpcore/issues/3737

[noissue]


mdellweg (approved):

ggainey:
lowerbounds will infinite-loop until https://github.com/pulp/pulpcore/pull/3924 is available - and after that, it will just **fail** because the fix that's being tested here, isn't in core/3.23, which is the current lowerbound for pulp_file/main.

Setting this to DRAFT - will work to get a "skip if server.status.core in (3.23/24/25)" (or something like that) in place.

lubosmj:
I am changing the pulpcore requirement here: https://github.com/pulp/pulp_file/pull/990/files#diff-4d7c51b1efe9043e44439a949dfd92e5827321b34082903477fd04876edb7552R1.

PR #970, opened 2023-06-14
Emit less warnings in tests

mdellweg:
[noissue]


lubosmj (approved):
`code` is deprecated since 3.1

https://docs.aiohttp.org/en/stable/client_reference.html#aiohttp.ClientResponseError.code

PR #990, opened 2023-06-30
Remove pulp-smash utils

### LINKED ISSUE #804: Migrate off of pulp-smash

lubosmj:
- [x] Clean up the code
- [x] Export some of the utilities to pulpcore

closes #804


pulp_file/tests/functional/api/from_pulpcore/test_content_cache.py
```diff
@@ -17,9 +17,10 @@
 
 is_redis_connected = get_redis_status()
```
mdellweg, line 18: This is unused now?

pulp_file/tests/functional/__init__.py
```diff
@@ -229,6 +229,17 @@ def file_fixture_server(file_fixtures_root, gen_fixture_server):
...
+def gen_file_remote(file_fixture_server):
+    def _gen_file_remote(manifest_path, **kwargs):
+        url = file_fixture_server.make_url(manifest_path)
+        data = {"url": url, "name": str(uuid.uuid4())}
+        data.update(kwargs)
+        return data
+
+    return _gen_file_remote
+
+
```
mdellweg, line 242: Do we really need to have this extra? Can we fold it into the `file_remote_factory`?

pulp_file/tests/functional/api/from_pulpcore/test_content_delivery.py
```diff
@@ -4,17 +4,14 @@
...
-)
-
 from pulpcore.client.pulp_file import (
     RepositorySyncURL,
 )
 
 from pulp_file.tests.functional.utils import (
     get_files_in_manifest,
     download_file,
+    gen_distribution,
```
mdellweg, line 14: As you can imagine, i'd love to see this just be part of the distribution factory.

pulp_file/tests/functional/api/from_pulpcore/test_content_cache.py
lubosmj: Yes. It will be removed.

pulp_file/tests/functional/__init__.py
lubosmj: This is a helper function for generating the request's body. In a couple of tests, we check if the requests' bodies for distributions or remotes were correctly parsed and then saved in Pulp.

pulp_file/tests/functional/api/from_pulpcore/test_content_delivery.py
lubosmj: Will try to think of a way to incorporate some parts of these helper functions inside the fixtures.

CHANGES/804.misc
```diff
@@ -0,0 +1 @@
+Removed pulp-smash utilities.
```
mdellweg, line 1:
```suggestion
Removed pulp-smash utilities for testing.
```
I suspect this little bit more context will save future us from search what pulp-smash is/was.

pulp_file/tests/functional/__init__.py
```diff
@@ -229,6 +229,20 @@ def file_fixture_server(file_fixtures_root, gen_fixture_server):
...
+@pytest.fixture
+def gen_file_remote(file_fixture_server):
+    def _gen_file_remote(manifest_path=None, **kwargs):
+        url = kwargs.get("url")
+        if not url:
+            assert manifest_path is not None
+            url = file_fixture_server.make_url(manifest_path)
+        data = {"url": url, "name": str(uuid.uuid4())}
+        data.update(kwargs)
+        return data
```
mdellweg, line 241: It's fine to "adjust" kwargs:
```suggestion
        if not url in kwargs:
            assert manifest_path is not None
            kwargs["url"] = file_fixture_server.make_url(manifest_path)
        kwargs.set_default("name", str(uuid.uuid4()))
        return kwargs
```

Also let's move this logic into the file_remote_factory.

pulp_file/tests/functional/api/test_rbac.py
```diff
@@ -238,33 +231,58 @@ def test_acs_apis(
...
+        file_distribution_api_client,
+        "create",
+        403,
+        {
+            "repository": admin_repo.pulp_href,
+            "name": str(uuid.uuid4()),
+            "base_path": str(uuid.uuid4()),
+        },
+    )
+    dis = try_action(
```
mdellweg, line 263: Can't we use the factories here? (By rewriting try_action, obviously...)
They would also give us proper cleanup in case a test fails.

pulp_file/tests/functional/api/from_pulpcore/test_distributions.py
```diff
@@ -123,52 +123,57 @@ def _update_distribution_and_assert(client, distribution_href, data):
...
-    distribution_data = gen_distribution(base_path=str(uuid4()).replace("-", "/"))
-    distribution = gen_object_with_cleanup(file_distribution_api_client, distribution_data)
+    distribution = file_distribution_factory(base_path=str(uuid4()).replace("-", "/"))
 
     # Test that spaces can not be part of ``base_path``.
     _create_distribution_and_assert(
-        file_distribution_api_client, gen_distribution(base_path=str(uuid4()).replace("-", " "))
+        file_distribution_api_client,
+        {"base_path": str(uuid4()).replace("-", " "), "name": str(uuid4())},
     )
```
mdellweg, line 135: Can we use the factory instead (even if it unexpectedly succeeds creating the distribution we get the cleanup code for free.)?
Also let's not move the assertion out to a subroutine.

pulp_file/tests/functional/api/test_rbac.py
lubosmj: The use of factories might overcomplicate things. I am leaving it as it is because `try_action` is currently used for different types of actions, such as update, create, read, or partial_update, and making it robust to consume factories may not be worth the effort. The PR is already enormous. I do not want to clutter it more. Are you okay with it?

mdellweg: I'll take that as a promise that we keep looking at these tests...

Idea for later: Maybe something like
```python
with bob, expect_code(403):
    file_remote_api_client.list()
```

pulp_file/tests/functional/api/test_sync.py
```diff
@@ -3,31 +3,25 @@
...
+    file_repository_version_api_client,
     file_remote_api_client,
     gen_object_with_cleanup,
     monitor_task,
+    fixtures_cfg,
+    wget_recursive_download_on_host,
 ):
     """Test syncing from a file repository with the file:// protocol handler"""
-    wget_download_on_host(FILE_FIXTURE_URL, "/tmp")
+    wget_recursive_download_on_host(urljoin(fixtures_cfg.remote_fixtures_origin, "file/"), "/tmp")
```
mdellweg, line 24: This is the thing we'd use the pulpcore pr for, right?
I don't know it being used by any other plugin. What if we just keep it here too and not bump the pulpcore version now?

lubosmj: Sure, that makes sense for now because of the bumping.

It is used in pulp_rpm though. We can move the fixture to pulpcore later as needed: https://github.com/search?q=org%3Apulp%20wget_download_on_host&type=code

lubosmj: Oh, the PR in pulpcore was already merged. I am going to copy then.

lubosmj:
The `pulp_content_url` fixture was introduced in pulpcore 3.27. I am using it in the tests after migrating off of pulpcore. I bumped the requirements because of that. I think we cannot backport it further down to 3.23.

gerrod3 (approved):

PR #1057, opened 2023-10-11
Add a Header ContentGuard test.

decko:
[noissue]

### DIFF
--- a/pulp_file/tests/functional/api/from_pulpcore/test_content_guard.py
+++ b/pulp_file/tests/functional/api/from_pulpcore/test_content_guard.py
@@ -1,6 +1,9 @@
 from aiohttp import BasicAuth
+import json
 import pytest
 import uuid
+
+from base64 import b64encode
 
 from pulpcore.client.pulp_file import (
     PatchedfileFileDistribution,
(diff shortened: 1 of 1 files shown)


decko:
> Why can't we add this test to the repository bringing the content guard?

The repository containing the content guard have unittests for it.
For functional tests we're only able to add it here given the components needed to run setup the tests.

mdellweg:
> > Why can't we add this test to the repository bringing the content guard?
> 
> The repository containing the content guard have unittests for it. For functional tests we're only able to add it here given the components needed to run setup the tests.

What components? pulp_file (released) is a test requirement for pulpcore what else is missing?

decko:
> > > Why can't we add this test to the repository bringing the content guard?
> > 
> > 
> > The repository containing the content guard have unittests for it. For functional tests we're only able to add it here given the components needed to run setup the tests.
> 
> What components? pulp_file (released) is a test requirement for pulpcore what else is missing?
> 
> There was a time when we had a rule that pulpcore tests must not rely on pulp_file, but that was when we tested main branches against each other. This is no longer true.

Wow! Thanks man. Shame on me, I didn't know that.
I've moved this test to `pulpcore` repo, where it belongs.

Thanks again @mdellweg

decko:
Closing given we found a better place to this piece of code.

PR #1068, opened 2023-10-26
Fix S3 CI runner

gerrod3:
[noissue]

### DIFF
--- a/pulp_file/tests/functional/api/from_pulpcore/test_content_cache.py
+++ b/pulp_file/tests/functional/api/from_pulpcore/test_content_cache.py
@@ -27,7 +27,7 @@ def test_full_workflow(
     pulp_content_url,
 ):
     if not redis_status:
-        pytest.skip("Could not connect to the Redis server")
+        pytest.xfail("Could not connect to the Redis server")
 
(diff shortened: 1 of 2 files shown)


gerrod3:
From my investigation I believe that S3 runner was getting stuck because of the pytest parallel workers failing to move onto the next test when using `pytest.skip` inside of a test. It happens to be that the only time this occurs is in `test_content_cache::test_full_workflow` when Redis is disabled which only occurs for the S3 runner. In my testing I couldn't get it to consistently reproduce, but I was able to confirm in every instance with a timeout it was because the pytest runner that executed that cache test failed to finish the next test it picked up. From my tests I found that adding the skip inside of a parallel marked test increased the chance that the pytest worker that ran that test would get stuck from running the next test. Using the skip decorator or module level skip had no impact on the pytest runner getting stuck. Also from my limited testing (4 separate runs), using `pytest.xfail` seems to fix the problem. 

I suggest we merge this as is and monitor the results. If it fails again we can modify the test to pass even without caching or move it to the serial tests.

ggainey (approved):

PR #1159, opened 2024-05-13
Update supported branches

mdellweg:
[noissue]


template_config.yml
```diff
@@ -58,10 +58,9 @@ stalebot_days_until_close: 30
 stalebot_days_until_stale: 90
 stalebot_limit_to_pulls: true
 supported_release_branches:
-- '1.10'
 - '1.11'
 - '1.12'
-- '1.13'
+- '1.14'
```
ggainey, line 63: We don't need to support 1.14

mdellweg:
Disabled the docs test. It was failing and we don't really care.

ggainey (approved):

